PR #26, opened 2018-01-12
[feature] Contributions

cazala:
List contributions on the sidebar


webapp/src/components/HomePage/Sidebar/ExpandedSidebar/WalletContributions/ContributionRow/ContributionRow.js
```diff
@@ -0,0 +1,23 @@
...
+    const { contribution } = this.props
+
+    return (
+      <div className="table-row">
+        <div className="col col-district">
+          <DistrictLink contribution={contribution} />
+        </div>
+        <div className="col col-contributed">
+          {contribution.land_count.toLocaleString()} LAND ({contribution.land_count *
+            1000}{' '}
```
NicoSantangelo: What do think about adding an object to handle this type of conversions? Maybe we could have a `land` domain similar to what we had here: https://github.com/decentraland/auction/blob/master/webapp/src/lib/land.js

so we can do `land.convert(contribution.land_count, 'mana')`

webapp/src/components/HomePage/Sidebar/ExpandedSidebar/WalletParcels/ParcelTable/ParcelTable.js
```diff
@@ -26,9 +26,7 @@ class ParcelTable extends React.PureComponent {
         </div>
       )
     } else {
-      return (
-        <div className="table-row-empty">You have no available land yet</div>
-      )
+      return <div className="table-row-empty">{"You don't own any land."}</div>
```
NicoSantangelo: Why `{}` here? ( for `{"You don't own any land."}` )

webapp/src/modules/address/sagas.js
```diff
@@ -38,3 +45,31 @@ function* handleAddressParcelsSuccess(action) {
     parcels: action.parcels
   })
 }
+
+function* handleAddressContributionsRequest(action) {
+  try {
+    const { address } = action
+    let contributions = yield call(() => api.fetchAddressContributions(address))
+
+    contributions = [
```
NicoSantangelo: I think we should remove this before merging into `master` don't we?

webapp/src/components/HomePage/Sidebar/ExpandedSidebar/WalletContributions/WalletContributions.js
```diff
@@ -0,0 +1,45 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+import ContributionTable from './ContributionTable'
+import Loading from 'components/Loading'
```
NicoSantangelo: Unused var!

NicoSantangelo:
screen shot 2018-01-14 at 12 49 04

What do you think about adding a third column `proposal` with the link to the github proposal?

abarmat:
Maybe some icon with the link to the proposal?

cazala:
like dis?
screen shot 2018-01-14 at 9 10 34 pm

webapp/src/components/HomePage/Sidebar/ExpandedSidebar/WalletParcels/ParcelTable/ParcelTable.js
cazala: it was cos of the single quote :P but yea I can alwas `&apos;`

PR #28, opened 2018-01-15
[feature] Show parcel name and description

cazala:
Show parcel name and description in popup. 

I also added two dictionaries to the wallet selector (`parcelsById` and `contributionsById`) to speed up the lookup in `getParcelAttributes` (used a lot to render the map), otherwise it will get slow specially for people with 100s or 1000s of parcels

### DIFF
--- a/webapp/src/lib/parcelUtils.js
+++ b/webapp/src/lib/parcelUtils.js
@@ -1,3 +1,5 @@
+import { shortenAddress } from 'lib/utils'
+
 export const ROADS_ID = 'f77140f9-c7b4-4787-89c9-9fa0e219b079'
 export const PLAZA_ID = '55327350-d9f0-4cae-b0f3-8745a0431099'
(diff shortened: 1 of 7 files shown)


webapp/src/lib/parcelUtils.js
```diff
@@ -70,23 +72,23 @@ export function getParcelAttributes(wallet, parcel, district) {
...
       }
     }
-    // if (district.name === 'Embassy Town') {
-    //   return {
-    //     label: 'District',
-    //     color: 'black',
-    //     backgroundColor: colors.DISTRICTS_CONTRIBUTED
-    //   }
-    // }
+    if (district && wallet.contributionsById[district.id]) {
```
NicoSantangelo: `wallet.contributionsById` 💯

NicoSantangelo:
🎉

PR #30, opened 2018-01-15
Update Front-End Setup README

cazala:
Update Front-End Setup README


yanis7774:
It will be great, if you will add to readme more information

cazala:
sorry, it was the wrong path 😅

PR #34, opened 2018-01-23
Feature/transfer land

NicoSantangelo:


webapp/src/modules/transfer/reducer.js
```diff
@@ -0,0 +1,48 @@
...
+  TRANSFER_PARCEL_SUCCESS,
+  TRANSFER_PARCEL_FAILURE,
+  CLEAN_TRANSFER
+} from './actions'
+
+const INITIAL_STATE = {
+  data: {},
+  loading: false,
+  error: null
+}
```
NicoSantangelo: _Reviewer_: I have a few doubts regarding the `transfer` domain.

The two main things are:
- It stands as a top level domain. My first instinct was to place it under parcels, but that also seems weird to me. I'm not sure where it's best (but it's easy enough to move around). What do you think?
- The structure of the data is:
```javascript
{
  hash: string,
  oldOwner: string,
  newOwner: string
}
```
which has the problem of having to `clean`  the transfer ( [see here](https://github.com/decentraland/marketplace/pull/34/files#diff-3157d100f0877a789f687b803efc6ffeR67) ) after closing the modal that creates it. Another option may be to have an object `{ 'x,y': { current data } }` as we use with parcels, but it seemed unnecessarily complex. Let me know if you have a better option

NicoSantangelo:
On top of the (already discussed) FontAwesome icons I would like to change this:

image

to a tooltip like https://kazzkiq.github.io/balloon.css/

But I don't know our stance on tooltips @abarmat @cazala

cazala:
@NicoSantangelo I added a bunch of stuff:

* Reflected successful transaction on the app state + UI, for this I had to [get rid of the request to fetch address' parcels](https://github.com/decentraland/marketplace/pull/34/commits/ea89509ee320597a89a36aeaa3fcc98061772e68#diff-239179f9180232a565d04034df6c0641L43), because that is done before the transaction is mined, so it returns the parcel you just transferred as yours, and overrides the app state.

* Batch calls to `ownerOfLandMany` because it was throwing due to a content-length too large. 

* I also skipped a broken test (it was broken before I started 😇 ), I made a few changes to the modal (made the close button more noticeable, and added a close CTA button on the transaction successful screen) and fixed a bunch of bugs around showing the owner in the popup and rendering owned/unowned land (it was depending on `price` instead of `owner` to tell the difference, and some parcels were rendered wrong).

webapp/src/modules/address/reducer.js
```diff
@@ -9,15 +9,21 @@ import {
...
+}
+
 const INITIAL_STATE = {
   data: {},
   loading: true,
   error: null
 }
-
 export default function reducer(state = INITIAL_STATE, action) {
+  const { TRANSFER_PARCEL_SUCCESS } = require('modules/transfer/actions')
```
NicoSantangelo: whops, why's this here?

cazala: oh yea, this was a workaround due to a cyclic import. if i import that at the top of the document is undefined. any better ideas to solve this?

NicoSantangelo:
- I un-commented the test as it seems to be working, I get `22 passing` tests without any change. Might have to do with this https://github.com/decentraland/marketplace/pull/34/commits/63a9dc0ee5f0a983ca5166fc152f6845927eceba#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R50 ?

Everything else looks 💯 but I'm not sure about the CLOSE button below the successfull transaction:
image

The text+button looks a bit cramped, no? 
Mayybe something like:

image

or

image

webapp/src/modules/address/reducer.js
NicoSantangelo: Aw, at the top of my head...not really. But I think we should leave a comment explaining the situation, no?

cazala:
regarding the test failing, this is what I get:

```
21 passing (593ms)
  1 failing

  1) ParcelService
       #addDbData
         should add the data stored on the db to each parcel and return a new array:
     error: column "name" of relation "parcels" does not exist
```

Maybe my local database is outdated? is that `column "name" of relation "parcels" does not exist` a postgres error?

cazala:
> image
>
> or
>
> image

Yep I like those way better! I think I am more inclined towards the second one but any of those work for me

NicoSantangelo:
> Maybe my local database is outdated? is that column "name" of relation "parcels" does not exist a postgres error?

Yes! Your `marketplace_test` database is outdated. Still, this is a problem (to tackle on another issue!) because it'll happen every-time the database changes, we're only adding columns when creating the tables.
A quick solution could be either:
- Dropping the `parcels` and re-running the tests
- Dropping the database, creating it again and re-running the tests
- Adding the `name` and `description` columns to the test database

Modal updated!

NicoSantangelo:
I can't approve "my own" changes but I think this is ready to be merged 🚀

PR #36, opened 2018-01-25
[fix] Show loader if there are pending requests to GET /parcels

cazala:
This keeps the decentraland logo in loader mode if there are pending requests to `GET /parcels`, before this, if the user dragged the map for a while, say, for 4 times, 4 requests would be sent, but as soon as the first request comes back that would unset the loader.

I changed the boolean `loading` prop to `loading_count` and made each `FETCH_PARCELS_REQUEST` increment it, and each `FETCH_PARCELS_FAILURE` or `FETCH_PARCELS_SUCCESS` decrement it.

We were using `FETCH_PARCELS_SUCCESS` to merge the addresses returned by `/addresses/:address/parcels` into the state, but this was screwing up the `loading_count`, so I added a new `MERGE_PARCELS` action that merges them into the app state without touching the count.


NicoSantangelo:
👍 worked really well

PR #39, opened 2018-01-26
Add metadata support onchain on the frontend and the backend.

NicoSantangelo:
Removes name and description from the db.
Adds asyncBatch as a helper.

Maybe related: https://github.com/decentraland/commons/issues/27

### DIFF
--- a/src/Parcel/Parcel.service.js
+++ b/src/Parcel/Parcel.service.js
@@ -1,4 +1,5 @@
 import { eth, Contract, Log } from 'decentraland-commons'
+import { LANDRegistry } from 'decentraland-commons/dist/contracts/LANDRegistry'
 
 import { Parcel } from './Parcel'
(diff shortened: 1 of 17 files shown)


abarmat:
Works as expected! What do you think if when we edit we complete the current information of the parcel?

edit

NicoSantangelo:
Good catch! I had that in mind but I forgot to add a `.data` to the state https://github.com/decentraland/marketplace/pull/39/files#diff-e5c30c1be25de34d40b365a6d64c1ad8R19

Testing the PR to re-add this feature, I realized that it was missing a bunch of UX. It worked but it didn't even have a `cancel` button. So I added a few more things. It works the same way as before but hopefully the experience is a bit better.
Let me know what you think (:

NicoSantangelo:
Rebasing!

PR #41, opened 2018-01-29
Toast messages

NicoSantangelo:
image

Not implemented as a result of any action, waiting for https://github.com/decentraland/marketplace/pull/39

### DIFF
--- a/webapp/src/components/HomePage/HomePage.js
+++ b/webapp/src/components/HomePage/HomePage.js
@@ -7,7 +7,7 @@ import Sidebar from './Sidebar'
 import MapComponent from './Map'
 import Minimap from './Minimap'
 import Modal from 'components/Modal'
-import { Toast } from 'components/Toast'
+import Toast from 'components/Toast'
 
 import './HomePage.css'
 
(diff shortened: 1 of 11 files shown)


NicoSantangelo:
Because we don't have any toasts yet, one way to try this locally could be:

**HomePage.container.js**

```javascript
import { connect } from 'react-redux'
import { openModal, openInfoToast } from 'modules/ui/actions'
import { fetchWallet } from 'modules/wallet/actions'

import HomePage from './HomePage'

const mapState = state => {
  return {
    isReady: true
  }
}

const mapDispatch = dispatch => ({
  onConnect: () => dispatch(fetchWallet()),
  onFirstVisit: () => dispatch(openModal('TermsModal')),
  onToast: () => dispatch(openInfoToast('Info toast')) // <---- here
})

export default connect(mapState, mapDispatch)(HomePage)
```

**HomePage.js**

```javascript
export default class HomePage extends React.PureComponent {
  static propTypes = {
    isReady: PropTypes.bool,
    onConnect: PropTypes.func
  }

  componentWillMount() {
    const { onConnect, onFirstVisit, onToast } = this.props

    onConnect()

    if (!localStorage.getItem('seenTermsModal')) {
      onFirstVisit()
    }

    setTimeout(() => onToast(), 3000) // <----- here
  }

  // (...)
```

PR #46, opened 2018-02-02
[feature] Display name on popup

cazala:
In this PR:

* I added a new `hoverParcel` action that fetches the parcel's data
* I modified  `fetchParcelData` to work with `x` and `y` instead of `parcel`
* I tweaked `getParcelAttributes` to return the right data for the Popup and made the proper changes in the Popup component
* I did some refactoring around the event handler to match the standards

To the reviewer:

The easier way to test this, if you have land, is to give a name to your parcel and then hover it: you should see the name in the Popup


webapp/src/components/HomePage/Map/ParcelsMap/ParcelPopup/ParcelPopup.js
```diff
@@ -15,7 +15,10 @@ export default function ParcelPopup(props) {
         </div>
       }
       <div className="body">
-        <div className="text label">{label}</div>
+        {label ? <div className="text label">{label}</div> : null}
+        {description ? (
+          <div className="text description">{description}</div>
+        ) : null}
         {/*description && <div className="text description">{description}</div>*/}
```
NicoSantangelo: I think we can safely remove this comment! :)

webapp/src/modules/parcels/sagas.js
```diff
@@ -37,6 +46,57 @@ function* handleParcelsRequest(action) {
...
+}
+
+function* handleParcelDataRequest(action) {
+  try {
+    const { x, y } = action
+    if (!inBounds(x, y)) return
+
+    const parcels = yield select(getParcels)
+    const parcel = parcels[buildCoordinate(x, y)]
+    if (!parcel) return
```
NicoSantangelo: Question for ` if (!inBounds(x, y)) return` and `if (!parcel) return`. Even though `handleHoverParcel` smartly avoids calls to this method when it's not necessary, shouldn't we `throw` here anyways?

webapp/src/modules/ui/map/sagas.js
```diff
@@ -1,16 +1,25 @@
...
-  })
+  yield put(fetchParcels(nw, se))
+}
+
+function* handleHoverParcel(action) {
+  const { x, y } = action
+  if (!inBounds(x, y)) return
+  const parcels = yield select(getParcels)
+  const parcel = parcels[buildCoordinate(x, y)]
+  if (!parcel || !parcel.owner || parcel.data) return // only fetch data if the parcel is loaded and has an owner, but the land data has not been loaded yet
```
NicoSantangelo: Would it be too much of an overhead to re-fetch the data even if we already have it?
So you don't have to refresh to see changes in other parcels, no?

NicoSantangelo: Furthermore, doing some logging it seems like all parcels have an empty `data` object like this:
image

so my data it's not being fetched when I hover.
When I remove the `|| parcel.data)` to give it a try, I don't see the popup update:

parcel

cazala:
Thanks for the review @NicoSantangelo! Updated ☝️ !

abarmat:
When fetching a parcel with empty metadata I get this error on the server 

[ParcelService] Unknown version when trying to decode land data: "" (see https://github.com/decentraland/commons/blob/master/docs/land-data.md)

Maybe we shouldn't show an error in those cases to avoid filling our logs with warnings.

NicoSantangelo:
> Maybe we shouldn't show an error in those cases to avoid filling our logs with warnings

agreed, we can change the `log.warn` for a `log.debug`, which won't run if `NODE_ENV=production`

webapp/src/components/HomePage/Map/ParcelsMap/ParcelsMap.js
```diff
@@ -96,6 +97,20 @@ export default class ParcelsMap extends React.Component {
...
+    // check if the land data has changed in order to update the popup
+    if (this.popup && this.tileHovered) {
+      const { x, y } = this.tileHovered
+      const coords = buildCoordinate(x, y)
+      const nextParcel = nextProps.parcels[coords]
+      const currentParcel = this.props.parcels[coords]
+      const nextParcelData = (nextParcel && nextParcel.data) || null
+      const currentParcelData = (currentParcel && currentParcel.data) || null
+      const isDifferent = !isEqual(nextParcelData, currentParcelData)
+      if (isDifferent) {
```
NicoSantangelo: What do you think of extracting:

```javascript
+   if (this.popup && this.tileHovered) {
+      const { x, y } = this.tileHovered
 +      const coords = buildCoordinate(x, y)
 +      const nextParcel = nextProps.parcels[coords]
 +      const currentParcel = this.props.parcels[coords]
 +      const nextParcelData = (nextParcel && nextParcel.data) || null
 +      const currentParcelData = (currentParcel && currentParcel.data) || null
 +      const isDifferent = !isEqual(nextParcelData, currentParcelData)
```

to it's own method (maybe removing the need for a comment) ?

so we can say 

```javascript
if (this.popupShouldUpdate(/* nextProps.parcels or maybe nextParcel */)) {   // first name that came to mind, not necessarily the best
    this.renderPopup()
}
```

webapp/src/components/HomePage/Map/ParcelsMap/ParcelsMap.js
```diff
@@ -243,12 +258,14 @@ export default class ParcelsMap extends React.Component {
...
-  // Called by the Parcel Grid on each tile render
-  getTileAttributes = latlng => {
-    const { wallet, parcels, districts } = this.props
+  getTileAttributesFromLatLng = latlng => {
     const { x, y } = this.mapCoordinates.latLngToCartesian(latlng)
+    const { wallet, parcels, districts } = this.props
+    return this.getTileAttributes(x, y, wallet, parcels, districts)
+  }
+  // Called by the Parcel Grid on each tile render
+  getTileAttributes = (x, y, wallet, parcels, districts) => {
```
NicoSantangelo: correct me if I'm wrong, but now that we have `getTileAttributesFromLatLng` we can change the firm of `getTileAttributes` to be `getTileAttributesFromLatLng = (x, y) {` to avoid having to pass 5 arguments and removing the repeated `const { wallet, parcels, districts } = this.props` on top of each call

webapp/src/components/HomePage/Map/ParcelsMap/ParcelsMap.js
```diff
@@ -317,36 +334,41 @@ export default class ParcelsMap extends React.Component {
...
+    leafletPopup.setLatLng(latlng).addTo(this.map)
     this.popup = leafletPopup
 
+    const { wallet, parcels, districts } = this.props
+    this.renderPopup(x, y, wallet, parcels, districts)
+
     return leafletPopup
   }
 
+  renderPopup = (x, y, wallet, parcels, districts) => {
```
NicoSantangelo: Same comment as https://github.com/decentraland/marketplace/pull/46/commits/aaeffc39fdb8ca9c42bfd4494d24ac0fd026f7ec#r165927360

webapp/src/modules/parcels/sagas.js
```diff
@@ -75,11 +75,15 @@ function* handleParcelRequest(action) {
...
-    if (!inBounds(x, y)) return
+    if (!inBounds(x, y)) {
+      throw new Error(`Coords (${x}, ${y}) are outside of the valid bounds`)
+    }
 
     const parcels = yield select(getParcels)
     const parcel = parcels[buildCoordinate(x, y)]
-    if (!parcel) return
+    if (!parcel) {
+      throw new Error(`Parcel (${x}, ${y}) is not in the state:`, parcels)
```
NicoSantangelo: Careful here! as far as I know, the second argument on the Error constructor is ignored. Only the first `message` argument is used

image

webapp/src/modules/parcels/utils.js
```diff
@@ -1,6 +1,21 @@
...
+    ipns: null,
+    version: 0
+  }
+}
+
+export function toParcelObject(parcelsArray, allParcels) {
   return parcelsArray.reduce((map, parcel) => {
-    map[parcel.id] = parcel
+    const oldParcel = (allParcels && allParcels[parcel.id]) || emptyParcel
+    map[parcel.id] = { ...oldParcel, ...parcel }
```
NicoSantangelo: What do you think of simplifying this a bit:

```javascript
export function toParcelObject(parcelsArray, allParcels = {}) {
  return parcelsArray.reduce((map, parcel) => {
    const oldParcel = allParcels[parcel.id]
    map[parcel.id] = { ...emptyParcel, ...oldParcel, ...parcel }
    return map
  }, {})
}
```

I **don't** want to impose the implementation, just as a discussion

NicoSantangelo:
It's working great for the parcel `name`, but _I think_ it's not reloading the description:

parcel

webapp/src/components/HomePage/Map/ParcelsMap/ParcelsMap.js
```diff
@@ -95,6 +97,20 @@ export default class ParcelsMap extends React.Component {
...
+    // check if the land data has changed in order to update the popup
+    if (this.popup && this.tileHovered) {
+      const { x, y } = this.tileHovered
+      const coords = buildCoordinate(x, y)
+      const nextParcel = nextProps.parcels[coords]
+      const currentParcel = this.props.parcels[coords]
+      const nextParcelData = (nextParcel && nextParcel.data) || null
+      const currentParcelData = (currentParcel && currentParcel.data) || null
+      const isDifferent = !isEqual(nextParcelData, currentParcelData)
+      if (isDifferent) {
```
NicoSantangelo: What do you think of extracting:

```javascript
+   if (this.popup && this.tileHovered) {
+      const { x, y } = this.tileHovered
 +      const coords = buildCoordinate(x, y)
 +      const nextParcel = nextProps.parcels[coords]
 +      const currentParcel = this.props.parcels[coords]
 +      const nextParcelData = (nextParcel && nextParcel.data) || null
 +      const currentParcelData = (currentParcel && currentParcel.data) || null
 +      const isDifferent = !isEqual(nextParcelData, currentParcelData)
```

to it's own method (maybe removing the need for a comment) ?

so we can say 

```javascript
if (this.popupShouldUpdate(/* nextProps.parcels or maybe nextParcel */)) {   // first name that came to mind, not necessarily the best
    this.renderPopup()
}
```

webapp/src/components/HomePage/Map/ParcelsMap/ParcelsMap.js
```diff
@@ -242,15 +258,17 @@ export default class ParcelsMap extends React.Component {
...
-  // Called by the Parcel Grid on each tile render
-  getTileAttributes = latlng => {
-    const { wallet, parcels, districts } = this.props
+  getTileAttributesFromLatLng = latlng => {
     const { x, y } = this.mapCoordinates.latLngToCartesian(latlng)
+    const { wallet, parcels, districts } = this.props
+    return this.getTileAttributes(x, y, wallet, parcels, districts)
+  }
+  // Called by the Parcel Grid on each tile render
+  getTileAttributes = (x, y, wallet, parcels, districts) => {
```
NicoSantangelo: correct me if I'm wrong, but now that we have `getTileAttributesFromLatLng` we can change the firm of `getTileAttributes` to be `getTileAttributesFromLatLng = (x, y) {` to avoid having to pass 5 arguments and removing the repeated `const { wallet, parcels, districts } = this.props` on top of each call

webapp/src/components/HomePage/Map/ParcelsMap/ParcelsMap.js
```diff
@@ -297,41 +316,59 @@ export default class ParcelsMap extends React.Component {
...
+    leafletPopup.setLatLng(latlng).addTo(this.map)
     this.popup = leafletPopup
 
+    const { wallet, parcels, districts } = this.props
+    this.renderPopup(x, y, wallet, parcels, districts)
+
     return leafletPopup
   }
 
+  renderPopup = (x, y, wallet, parcels, districts) => {
```
NicoSantangelo: Same comment as https://github.com/decentraland/marketplace/pull/46/commits/aaeffc39fdb8ca9c42bfd4494d24ac0fd026f7ec#r165927360

webapp/src/modules/parcels/utils.js
NicoSantangelo: What do you think of simplifying this a bit:

```javascript
export function toParcelObject(parcelsArray, allParcels = {}) {
  return parcelsArray.reduce((map, parcel) => {
    const oldParcel = allParcels[parcel.id]
    map[parcel.id] = { ...emptyParcel, ...oldParcel, ...parcel }
    return map
  }, {})
}
```

I **don't** want to impose the implementation, just as a discussion

webapp/src/components/HomePage/Map/ParcelsMap/ParcelsMap.js
NicoSantangelo: What do you think of extracting:

```javascript
+   if (this.popup && this.tileHovered) {
+      const { x, y } = this.tileHovered
 +      const coords = buildCoordinate(x, y)
 +      const nextParcel = nextProps.parcels[coords]
 +      const currentParcel = this.props.parcels[coords]
 +      const nextParcelData = (nextParcel && nextParcel.data) || null
 +      const currentParcelData = (currentParcel && currentParcel.data) || null
 +      const isDifferent = !isEqual(nextParcelData, currentParcelData)
```

to it's own method (maybe removing the need for a comment) ?

so we can say 

```javascript
if (this.popupShouldUpdate(/* nextProps.parcels or maybe nextParcel */)) {   // first name that came to mind, not necessarily the best
    this.renderPopup()
}
```

webapp/src/components/HomePage/Map/ParcelsMap/ParcelsMap.js
NicoSantangelo: correct me if I'm wrong, but now that we have `getTileAttributesFromLatLng` we can change the firm of `getTileAttributes` to be `getTileAttributesFromLatLng = (x, y) {` to avoid having to pass 5 arguments and removing the repeated `const { wallet, parcels, districts } = this.props` on top of each call

webapp/src/components/HomePage/Map/ParcelsMap/ParcelsMap.js
NicoSantangelo: Same comment as https://github.com/decentraland/marketplace/pull/46/commits/aaeffc39fdb8ca9c42bfd4494d24ac0fd026f7ec#r165927360

webapp/src/modules/parcels/utils.js
NicoSantangelo: What do you think of simplifying this a bit:

```javascript
export function toParcelObject(parcelsArray, allParcels = {}) {
  return parcelsArray.reduce((map, parcel) => {
    const oldParcel = allParcels[parcel.id]
    map[parcel.id] = { ...emptyParcel, ...oldParcel, ...parcel }
    return map
  }, {})
}
```

I **don't** want to impose the implementation, just as a discussion

webapp/src/modules/parcels/sagas.js
```diff
@@ -37,6 +46,61 @@ function* handleParcelsRequest(action) {
...
+  try {
+    const { x, y } = action
+    if (!inBounds(x, y)) {
+      throw new Error(`Coords (${x}, ${y}) are outside of the valid bounds`)
+    }
+
+    const parcels = yield select(getParcels)
+    const parcel = parcels[buildCoordinate(x, y)]
+    if (!parcel) {
+      throw new Error(`Parcel (${x}, ${y}) is not in the state:`, parcels)
```
NicoSantangelo: Careful here! as far as I know, the second argument on the Error constructor is ignored. Only the first `message` argument is used

image

webapp/src/components/HomePage/Map/ParcelsMap/ParcelsMap.js
cazala: 1) The reason to have `getTileAttributes` and `getTileAttributesFromLatLng` is because `LeafletParcelGrid` only speaks `latlng`, while `ParcelsMap` uses `x, y`. The previous signature only used `latlng` but we were converting `x,y` to `latlng` all the time inside `ParcelsMap` which was converted back to `x,y` inside `getTileAttributes`, it seemed wasteful so I splitted it in two methods and avoided so much going back and forth between parcel coods and  latlng coords. 

2) I extracted  `wallet, parcels, districts` outside of `getTileAttributes` and `renderPopup` because they are not always coming from `this.props`. In `shouldComponentUpdate`,  those values have to be deconstructed from `nextProps` instead.

cazala:
@NicoSantangelo [regarding this](https://github.com/decentraland/marketplace/pull/46#issuecomment-363049880), yes, the popup was not refreshing when the wallet data was updated, which contains your address' parcels, used to determine if a parcel should say `Your parcels` or `Owner: 0x..`, it's fixed by  dde4c30de39cb336aac869d7759c6da8e0842466

@abarmat [regarding this](https://github.com/decentraland/marketplace/pull/46#issuecomment-362978444) I did what @NicoSantangelo [suggested here](https://github.com/decentraland/marketplace/pull/46#issuecomment-363036425), it's fixed by 3565ef53dc43ad784d77d04330a0df9d797c72b2

webapp/src/components/HomePage/Map/ParcelsMap/ParcelsMap.js
cazala: https://github.com/decentraland/marketplace/pull/46#discussion_r165978859

NicoSantangelo:
Awesome work!

PR #47, opened 2018-02-02
Fix the map render on negative coordinates

NicoSantangelo:
Steps to reproduce the bug:

- Load /-108/-82
- The map will move to a Y coordinate above 0
- Try to go to a Y coordinate below 0 by dragging the map


cazala:
purrrrfect 😸 !

PR #53, opened 2018-02-06
[chore] Refactor sagas with action creators

cazala:
Refactored sagas to use action creators instead of creating plain action objects according to the new [standards](https://github.com/decentraland/standards/blob/master/standards/react-redux.md#25-sagas)


NicoSantangelo:
Awesome work!

I have a few teeny tiny comments which are non-blocking so I'll DM you

PR #63, opened 2018-02-12
[feature] Publications

cazala:
WIP

### DIFF
--- a/webapp/src/modules/address/utils.js
+++ b/webapp/src/modules/address/utils.js
@@ -1,3 +1,19 @@
 export function toAddressParcelIds(parcelsArray) {
   return parcelsArray.map(parcel => parcel.id)
 }
+
+export function toAddressPublicationIds(publicationsArray) {
+  return publicationsArray.map(publication => publication.tx_hash)
+}
+
+export function grab(all, ids) {
+  const array = []
+  const byId = {}
+  ids.forEach(id => {
+    if (all[id]) {
+      array.push(all[id])
+      byId[id] = all[id]
+    }
+  })
+  return [array, byId]
+}
--- a/webapp/src/lib/api.js
+++ b/webapp/src/lib/api.js
@@ -1,9 +1,139 @@
 import axios from 'axios'
+import _sortBy from 'lodash/sortBy'
 
 import { env } from 'decentraland-commons'
 
 const httpClient = axios.create()
 const URL = env.get('REACT_APP_API_URL', '')
+const FILTER_DEFAULTS = {
+  limit: 20,
+  offset: 0,
+  sortBy: 'created_at',
+  sortOrder: 'asc'
+}
+
+const MOCK_PUBLICATIONS = [
+  {
+    tx_hash:
+      '0xf3dfe93e261e6be5ef615ff5da8b09bb4ad4212bfb03dd1b2e41ccafbefefa11',
+    address: '0x0de1ec708665b93478bd0ed264e60b68a8663bb4',
+    x: 106,
+    y: -84,
+    price: '10000',
+    tx_status: 'pending',
+    created_at: 1518474436085,
+    expires_at: 1518647236085,
+    sold: false
+  },
+  {
+    tx_hash:
+      '0xc36a170103da0ed4055125b8c39fd49af0efa6a5ee1576297126727bbda6cc42',
+    address: '0x0de1ec708665b93478bd0ed264e60b68a8663bb4',
+    x: 65,
+    y: 38,
+    price: '23000',
+    tx_status: 'confirmed',
+    created_at: 1518474436085,
+    expires_at: 1518647236085,
+    sold: false
+  },
+  {
+    tx_hash:
+      '0x5ff0b21b99581a8d86021dfd84e3c08eeffeb4d3490857173f532f8da2154266',
+    address: '0x0de1ec708665b93478bd0ed264e60b68a8663bb4',
+    x: 45,
+    y: 65,
+    price: '1000',
+    tx_status: 'failed',
+    created_at: 1518474436085,
+    expires_at: 1518647236085,
+    sold: false
+  },
+  {
+    tx_hash:
+      '0xbce357f29b2acb14f28d79d920fb9f4648158f5123b84e4b3e4045ef8d8fd7d9',
+    address: '0x0de1ec708665b93478bd0ed264e60b68a8663bb4',
+    x: 89,
+    y: 34,
+    price: '15000',
+    tx_status: 'confirmed',
+    created_at: 1518474436085,
+    expires_at: 1518647236085,
+    sold: true
+  },
+  {
+    tx_hash:
+      '0x09719587a39352398da2ef774af93fc8050ab33c63d75f7a71f9e12acb2481f2',
+    address: '0xec6e6c0841a2ba474e92bf42baf76bfe80e8657c',
+    x: 26,
+    y: 12,
+    price: '20000',
+    tx_status: 'confirmed',
+    created_at: 1518820036085,
+    expires_at: 1518992836085,
+    sold: false
+  },
+  {
+    tx_hash:
+      '0xf7b2def1afb9391e1fc0d99211fc0491888333559f28858182fbaf2ef2033800',
+    address: '0xec6e6c0841a2ba474e92bf42baf76bfe80e8657c',
+    x: 55,
+    y: 55,
+    price: '2000',
+    tx_status: 'confirmed',
+    created_at: 1519120036085,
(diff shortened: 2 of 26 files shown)


webapp/src/components/MarketplacePage/Publication/Publication.js
```diff
@@ -0,0 +1,70 @@
...
+          >
+            <EtherscanLink txHash={publication.tx_hash}>
+              <Icon name={iconName} className={publication.tx_status} />
+            </EtherscanLink>
+          </span>
+        </Card.Content>
+        <Card.Content extra>
+          <span className="footer">
+            <Header size="medium" floated="left" className="price">
+              {numeral(publication.price).format('0,0.00')} MANA
```
NicoSantangelo: I have two pet peeves here:

### 1 

I think it's better to show the entire number and use text-overflow with a tooltip if it's too long, to avoid confusion.
Furthermore, I think it's distracting to _always_ show decimals, like in `23.000,00 MANA`. I like `23.000 MANA` more

### 2

I'm not agains using `numeral` here but I think we'll need a better abstraction, as we are hardcoding US-based formatting. The rest of the codebase is using `toLocaleString` so is not doing a good job of fixing decimals, but at least we follow the browser locale.

For example we could have a `numberFormatter.js` (or just `numeral`) on `./lib` with one or more `format` methods so we can abstract number formatting in a single place.

---

What do you think?

webapp/src/components/MarketplacePage/MarketplacePage.js
```diff
@@ -0,0 +1,76 @@
...
+    text: 'Most expensive',
+    value: 'most-expensive'
+  },
+  {
+    text: 'Closest to expire',
+    value: 'closest-to-expired'
+  }
+]
+
+export default class ParcelDetailPage extends React.PureComponent {
```
NicoSantangelo: Whoops! this should be called `MarketplacePage`

webapp/src/components/MarketplacePage/Publication/Publication.js
```diff
@@ -0,0 +1,70 @@
...
+      iconTooltip = 'Transaction pending'
+    } else if (publication.tx_status === txUtils.TRANSACTION_STATUS.failed) {
+      iconName = 'remove circle outline'
+      iconTooltip = 'Transaction failed'
+    }
+    const isExpired = publication.expires_at < Date.now()
+    return (
+      <Card className="Publication">
+        <Card.Content>
+          <Link to={`/${publication.x}/${publication.y}`}>
```
NicoSantangelo: To avoid future route renaming pains, this should be using `locations.parcelMapDetail(publication.x, publication.y)`

webapp/src/components/MarketplacePage/Publication/Publication.js
```diff
@@ -0,0 +1,70 @@
...
+            <Button
+              floated="right"
+              size="tiny"
+              disabled={
+                isExpired ||
+                publication.sold ||
+                publication.tx_status !== txUtils.TRANSACTION_STATUS.confirmed
+              }
+            >
+              {publication.sold ? 'Sold' : 'Buy'}
```
NicoSantangelo: The server changed from `sold` to `is_sold` to show that this is a Boolean at first glance.

Sorry about that, I probably didn't warn you (you'll have to change the mock 😞)

webapp/src/components/MarketplacePage/Publication/Publication.js
```diff
@@ -0,0 +1,70 @@
+import React from 'react'
+import { Header, Card, Button, Icon } from 'semantic-ui-react'
+import { Link } from 'react-router-dom'
+import { publicationType } from 'components/types'
+import { txUtils } from 'decentraland-commons'
+import EtherscanLink from 'components/EtherscanLink'
+import numeral from 'numeral'
+import moment from 'moment'
+import './Publication.css'
```
NicoSantangelo: I know this is nitpicking, but can we keep the import order more or less simliar to `MarketplacePage.js`s ? (which is how most components order the imports

I think that instead of just pointing it out here I should add a PR to the standards, but in my head this makes sense:

```javascript
import React from 'react'
// React related imports, like react-router-dom or prop-types
// general imports, like decentraland-commons or lodash.debounce

// Components, semantic-ui-react first just to distinguish it
// Component/Types

// Lib or Utils

// CSS
```

I don't intend to always keep _that_ template, but I do think that grouping imports by "category" can be useful for taking a quick look.

webapp/src/components/MarketplacePage/Publication/Publication.js
```diff
@@ -0,0 +1,70 @@
...
+          <Card.Meta>
+            {isExpired
+              ? `Expired`
+              : `Expires ${moment(publication.expires_at).fromNow()}`}
+          </Card.Meta>
+          <span
+            data-balloon-pos="up"
+            data-balloon={iconTooltip}
+            className="tx-icon"
+          >
```
NicoSantangelo: _**Not necessarily actionable**_

This looks like a component we'll reuse later. We don't have to preemptively abstract it, but we might want to keep an eye on it

webapp/src/modules/address/utils.js
```diff
@@ -1,3 +1,19 @@
 export function toAddressParcelIds(parcelsArray) {
   return parcelsArray.map(parcel => parcel.id)
 }
+
+export function toAddressPublicationIds(publicationsArray) {
+  return publicationsArray.map(publication => publication.tx_hash)
+}
+
+export function grab(all, ids) {
```
NicoSantangelo: Can we search for a name a bit more descriptive here? I think it's a bit generic and it prompts you to search the implementation to see what it does.

It's a bit hard to name because it's doing two things at the same time, but maybe:

`pickAndMap()`

ugh

webapp/src/components/MarketplacePage/MarketplacePage.js
```diff
@@ -0,0 +1,172 @@
...
+    }
+  }
+
+  navigateTo(options) {
+    const navigationOptions = {
+      ...this.props,
+      ...options
+    }
+    const { page, sortBy, sortOrder, onNavigate } = navigationOptions
+    const url = `/marketplace?page=${page}&sort_by=${sortBy}&sort_order=${sortOrder}`
```
NicoSantangelo: Given that we have methods like `getPageFromRouter` that use `QueryString` on `./utils` can we do something like:

```javascript
import { buildFilterUrl } from './utils'

buildUrl(navigationOptions)

function buildUrl(navigationOptions) {
  return `${locations.marketplace}?${QueryString.stringify(navigationOptions)`
}
```

(not to be followed word-by-word, maybe unpacking is best to ensure certain params)

webapp/src/components/MarketplacePage/utils.js
```diff
@@ -0,0 +1,59 @@
+import QueryString from 'query-string'
```
NicoSantangelo: **Nitpick**, I think this should be `queryString`

webapp/src/components/ParcelDetailPage/ParcelDetailPage.js
```diff
@@ -2,8 +2,10 @@ import React from 'react'
 import PropTypes from 'prop-types'
 
 import { locations } from 'locations'
+import { Container, Header } from 'semantic-ui-react'
```
NicoSantangelo: `Header` is no used (lint error)

webapp/src/components/ParcelPreview/ParcelCanvas/ParcelCanvas.js
```diff
@@ -0,0 +1,142 @@
...
+    this.oldState = this.state
+  }
+
+  componentDidMount() {
+    this.renderMap()
+  }
+
+  renderMap() {
+    if (!this.canvas) {
+      return '🦄'
```
NicoSantangelo: 💯

NicoSantangelo:
@cazala I think we can remove the `// TODO: remove mock` as I'm adding a way to add fake data as a standalone script (:

webapp/src/components/MarketplacePage/MarketplacePage.js
cazala: I can't use `queryString.stringify` because the props and the query params use different casing: `sortBy` vs `sort_by`. But I can move that to utils and use `locations.marketplace` instead of `/marketplace`

cazala: or i can use a `camelCase` to `snake_case` helper + `queryString.stringify`

cazala:
@NicoSantangelo updated ☝️ ^^

PR #66, opened 2018-02-16
[feature] Publish LAND

NicoSantangelo:
continues #63


webapp/src/modules/publication/selectors.js
```diff
@@ -1,4 +1,25 @@
...
 export const isLoading = state => getState(state).loading.length > 0
 export const getError = state => getState(state).error
+
+export const getPublications = createSelector(
+  getData,
+  state => getTransactionsByType(state, PUBLISH_SUCCESS),
+  (publications = {}, publishTransactions) => {
+    for (const transaction of publishTransactions) {
+      const publication = transaction.payload
+      publications[publication.tx_hash] = {
```
cazala: Watch out, this is mutating the app state. The app state should be immutable and new states should only be generated from reducers. We should create a new object here and assign the enriched publications to it, then return that new object.

webapp/src/locations.js
```diff
@@ -11,14 +11,13 @@ export const locations = {
...
 
-  colorCodes: '/colorCodes',
-  privacy: '/privacy',
-
   parcel: '/:x/:y/detail',
   parcelDetail: (x, y) => `/${x}/${y}/detail`,
 
-  parcelTransfer: '/:x/:y/transfer',
-  parcelTransferDetail: (x, y) => `/${x}/${y}/transfer`,
+  activities: '/activities',
```
cazala: `/activities` reminds me of [step brothers](https://www.google.com.ar/search?q=so+much+room+for+activities&tbm=isch) :P can we go with `/activity`?

NicoSantangelo:
Updated!

PR #68, opened 2018-02-21
[feature] Parcel detail modifications

cazala:
Updated `ParcelDetailPage` to match designs, added a `Parcel` component to encapsulate some behaviour repeated across the parcel screens, and refactored those screens to use it. Fixed some bugs along the way.


NicoSantangelo:
Awesome!
I have some _really minor_ comments, so I'll DM you about them instead of delaying this merge.

Great work!

PR #69, opened 2018-02-21
[Feature] index blockchain

NicoSantangelo:
Indexes the blockchain and stores data on the database
Updates API accordingly
Updates Schema using new migrations model
Updates initialize scripts

~~Is missing the owner field from the `AuctionCreated` event on the [Marketplace contract](https://github.com/decentraland/marketplace-contracts/blob/master/contracts/marketplace/Marketplace.sol) @abarmat @chentschel~~
https://ropsten.etherscan.io/address/0x33a6a57de7320e4ab9c6452447a0a4b9eab0a704 🚀 

If you're coming from master and want to run the migrations you can update your migrations table with this:

```sql
INSERT INTO pgmigrations (
 name, run_on
) VALUES (
  '1519213221468_parcels-create',
  NOW()
), (
  '1519213556550_districts-create',
  NOW()
),  (
  '1519213885561_contributions-create',
  NOW()
),  (
  '1519214706892_publications-create',
  NOW()
)
```


scripts/monitor/handlers/index.js
```diff
@@ -1,10 +1,109 @@
...
 
-export function transform_LANDRegistry_Transfer(event) {
+export async function transform_LANDRegistry(event) {
+  if (event.removed) return
+
+  const { assetId } = event.args
+  const id = await decodeAssetId(assetId)
+
+  switch (event.event) {
+    case 'Update': {
```
cazala: Shouldn't this be `'Transfer'` and the case below `'Update'` instead?

cazala:
This is looking great! I can't wait to use use the map view after this ⚡️!

cazala:
Btw, could we add some db migration instructions on the readme? 😇

scripts/monitor/handlers/index.js
NicoSantangelo: Whoops good catch!

cazala:
Awesome work! 🥇

PR #71, opened 2018-02-23
[Feature] Parcel publications endpoint

NicoSantangelo:
Blocked by #70 (continuation)

@cazala What do you think of this: https://github.com/decentraland/marketplace/commit/514d8cb9cea87d4ac340282b102ede5863c8e3e7
I removed both the parcel data endpoint _and_ the hover event in an attempt to make things simpler, but that means no more "realtime" data on the server. You'll need to refresh to get the latest data.

### DIFF
--- a/src/Parcel/Parcel.js
+++ b/src/Parcel/Parcel.js
@@ -54,10 +54,10 @@ export class Parcel extends Model {
     const [maxx, maxy] = coordinates.toArray(max)
 
     return await this.db.query(
-      `SELECT par.*, row_to_json(pub.*) as publication
+      `SELECT DISTINCT ON(par.id, pub.status) par.*, row_to_json(pub.*) as publication
         FROM ${this.tableName} as par
         LEFT JOIN (
-          ${Publication.findLastOpenSql()}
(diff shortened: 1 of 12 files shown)


cazala:
> I removed both the parcel data endpoint and the hover event in an attempt to make things simpler, but that means no more "realtime" data on the server. You'll need to refresh to get the latest data.

I think that's fine 👍. Dragging the map would refresh the parcel data for the whole window anyway, right?

NicoSantangelo:
> Dragging the map would refresh the parcel data for the whole window anyway, right?

Yes!
Rebased.

cazala:
@NicoSantangelo the endpont `/api/parcels/:x/:y/publications` is working fine for me, but the `/api/parcels?nw&se` is always returning `publication: null` even for parcels that return publications by hitting the other endpoint, like this:

publications-endpoint

NicoSantangelo:
Didn't see the comment! let me check

NicoSantangelo:
@cazala Hmm it looks like it's working on my end

image

are you sure you have a `confirmed` and `open` publication for your parcel?

PR #72, opened 2018-02-25
[feature] Profile Page

cazala:
Adds a profile page that can be accessed from `/address/:address/:tab`, it shows user's parcels, contributions and publications.

profile page


NicoSantangelo:
A few comments:

**1)** Got a few prop-type errors on the console:

```
Failed prop type: The prop `sortBy` is marked as required in `ProfilePage`, but its value is `undefined`
Failed prop type: The prop `sortOrder` is marked as required in `ProfilePage`, but its value is `undefined`
Failed prop type: The prop `onFetchPublications` is marked as required in `ProfilePage`, but its value is `undefined`

Warning: Failed prop type: Invalid prop `color` of value `gray` supplied to `Label`, expected one of ["red","orange","yellow","olive","green","teal","blue","violet","purple","pink","brown","grey","black"].
```

**2)** Can we hide the pagination when we only have the first one?

image

**3)** I know we don't have much of a choice, but loading the list requests a range for each parcel

image

and it refetches everything if you go back and forth between the tabs. I don't think that optimizing this should be part of this issue, but we may wan't to think about it for later.

src/Parcel/Parcel.service.js
```diff
@@ -58,7 +58,7 @@ export class ParcelService {
       let data
 
       try {
-        const landData = await contract.landData(parcel.x, parcel.y)
+        const landData = await contract.getData(parcel.x, parcel.y)
```
NicoSantangelo: This is ok, but keep in mind that ultimately, this should call `landData` (like it does on master). To avoid having a breaking merge, do you want to rebase master?

webapp/src/components/AddressLink/AddressLink.js
```diff
@@ -15,29 +15,39 @@ export default class AddressLink extends React.Component {
...
+      scale,
+      hasTooltip,
+      isUser,
+      className
+    } = this.props
 
     return (
       <div
-        className="AddressLink"
+        className={`AddressLink ${className}`}
```
NicoSantangelo: Shouldn't we remove this Component from the `HomePage.js` page?

webapp/src/components/ProfilePage/Contribution/Contribution.js
```diff
@@ -0,0 +1,46 @@
+import React from 'react'
+import { Card, Header, Button } from 'semantic-ui-react'
+import { contributionType } from 'components/types'
+
+import './Contribution.css'
+
+export default class Publication extends React.PureComponent {
```
NicoSantangelo: Whoops Component name

webapp/src/components/ProfilePage/Contribution/Contribution.js
```diff
@@ -0,0 +1,46 @@
...
+          <span className="footer">
+            <Header size="medium" floated="left" className="land-count">
+              <span className="amount">{contribution.land_count}</span>{' '}
+              &nbsp;LAND
+            </Header>
+            <a
+              href={contribution.district ? contribution.district.link : '#'}
+              target="_blank"
+              rel="noopener noreferrer"
+            >
```
NicoSantangelo: I think this may be confusing. Even if the contribution says loading, clicking a link and getting a `#` it's never nice.

What do you think about doing something like:

```javascript
{contribution.district ? (
  <a
    href={contribution.district.link}
    target="_blank"
    rel="noopener noreferrer"
  >
    <Button floated="right" size="tiny">
      Proposal
    </Button>
  </a>
) : (
  <Button floated="right" size="tiny" disabled={true}>
    Proposal
  </Button>
)}
```

(not tested, wrote here)

webapp/src/components/ProfilePage/Parcel/Parcel.js
```diff
@@ -0,0 +1,37 @@
...
+
+import { locations } from 'locations'
+import { Card } from 'semantic-ui-react'
+import ParcelName from 'components/ParcelName'
+import ParcelPreview from 'components/ParcelPreview'
+import { parcelType } from 'components/types'
+
+import './Parcel.css'
+
+export default class Publication extends React.PureComponent {
```
NicoSantangelo: Whoops Component name

webapp/src/components/ProfilePage/ProfilePage.js
```diff
@@ -0,0 +1,208 @@
...
+      page: data.activePage,
+      address,
+      tab
+    })
+    onNavigate(url)
+  }
+
+  renderLoading() {
+    return (
+      <Dimmer active inverted>
```
NicoSantangelo: I think we can safely remove the `Dimmer` here, because the loading is inside the Tab content so it looks the same (and I actually like it that way, dimming can be a bit annoying)

webapp/src/components/ProfilePage/ProfilePage.js
```diff
@@ -0,0 +1,208 @@
...
+            {grid.map(parcel => <Parcel key={parcel.id} parcel={parcel} />)}
+          </Card.Group>
+        )
+      }
+      case TABS.CONTRIBUTIONS: {
+        return (
+          <Card.Group stackable={true}>
+            {grid.map(contribution => (
+              <Contribution
+                key={contribution.district_id}
```
NicoSantangelo: Sadly `district_id` can be duplicated because an address could've contributed to the same district multiple times for different `land_count`s (and we're not merging those anywhere). Maybe we can use `index` here?

Also, to discuss later, we might want to add the contribution date to the UI.

webapp/src/lib/LeafletParcelGrid.js
```diff
@@ -169,14 +171,24 @@ export const LeafletParcelGrid = L.Layer.extend({
     // Clear canvas
     const ctx = this.canvas.getContext('2d')
     ctx.clearRect(0, 0, this.canvas.width, this.canvas.height)
-
-    for (let index = tiles.length - 1; index >= 0; index--) {
-      this.renderTile(tiles[index])
+    let marker = null
+    for (let index = 0; index < tiles.length; index++) {
+      const point = this.renderTile(tiles[index])
+      if (point) {
```
NicoSantangelo: _To consider_

I think renderTile is doing two things at the same time and it's a bit confusing. I wouldn't have guessed that the return value signifies that a marker is present.
I know it's a bit less performant but my first instinct is to do something like:

```javascript
for (let index = 0; index < tiles.length; index++) {
  const tile = tiles[index]
  const tileId = this.renderTile(tile)
  if (this.options.marker === tileId) {
    marker = this.map.latLngToContainerPoint(tile.center)
  }
// (...)

renderTile(tile) {
  // (...)
  return id
},
```

What do you think?

webapp/src/lib/parcelUtils.js
```diff
@@ -121,3 +121,50 @@ export function getParcelAttributes(wallet, parcel, districts) {
     backgroundColor: COLORS.taken
   }
 }
+
+export function drawMarker(
```
NicoSantangelo: _To consider_

I think this might deserve to be on it's own file. Like:

```javascript
import { marker } from './marker'

marker.draw(/* (...) */)
```

What do you think?

webapp/src/locations.js
```diff
@@ -1,11 +1,12 @@
 export const locations = {
   root: '/',
 
-  address: '/address/:address',
-  addressDetail: address => `/address/${address}`,
+  profile: '/address/:address/:tab',
+  profilePage: (address, tab = 'parcels') => `/address/${address}/${tab}`,
```
NicoSantangelo: Ugh, I think we should use `TABS` here but I don't want locations to import external objects 😓 ...Maybe have a `export const PROFILE_PAGE_TABS = {}` here, and use that in the Component?

webapp/src/components/ProfilePage/ProfilePage.js
cazala: This is a pattern enforced by Semantic UI, `<Loader>` has a `display: none;` by default unless it's inside a `<Dimmer>`

webapp/src/components/ProfilePage/ProfilePage.js
cazala: contributions are getting grouped by `district_id`. I added two contributions of 6 LAND to `AETHERIAN project`  and the API returns a single contribution of 12 LAND

webapp/src/components/AddressLink/AddressLink.js
cazala: I will remove it on the next PR, where I get rid of the Sidebar as well and I introduce the new top navbar (;

cazala:
@NicoSantangelo thanks for the review! I addressed most of the change requests, and on the ones I didn't I left a comment explaining why 🦄

webapp/src/components/ProfilePage/ProfilePage.js
cazala: 👉 https://github.com/decentraland/marketplace/blob/master/src/Contribution/Contribution.js#L22

webapp/src/components/ProfilePage/ProfilePage.js
cazala: > Loaders are hidden unless has prop `active` or inside an `Dimmer active`.

my bad 🤦‍♂️ updated!

NicoSantangelo:
Super awesome

PR #78, opened 2018-03-02
[Feature] Add persisten local storage for txs.

NicoSantangelo:
Also:
- Move middlewares to their own modules.
- Watch pending tx on load


webapp/src/modules/transaction/sagas.js
```diff
@@ -2,30 +2,39 @@ import { takeEvery, put, call, select } from 'redux-saga/effects'
...
 export function* transactionSaga() {
   yield takeEvery(FETCH_TRANSACTION_REQUEST, handleTransactionRequest)
+  yield takeEvery(WATCH_LOADING_TRANSACTIONS, handleWatchLoadingTransactions)
+}
+
+const watchIndex = {
+  // hash: true
 }
 
 function* handleTransactionRequest(action = {}) {
```
NicoSantangelo: I think this might be confusing now. Because the action is `fetching` but the actual behaviour is watching. I don't know if it's necessary to change it now but advice is appreciated.

An example of what I'm thinking is:

```javascript
export function* transactionSaga() {
  yield takeEvery(WATCH_TRANSACTION_REQUEST, handleWatchTransactionRequest)
  yield takeEvery(WATCH_LOADING_TRANSACTIONS, handleWatchLoadingTransactions)
}

function* handleTransactionWatch(action = {}) {
  try {
    // (...)
    yield put(
      watchTransactionSuccess({
        ...transaction,
        ...confirmedTx
      })
    )
  } catch (error) {
    yield put(watchTransactionFailure(transaction, error.message))
  }
}

function* handleWatchLoadingTransactions(action) {
  const transactionRequests = yield select(getLoading)

  for (const action of transactionRequests) {
    const hash = getTransactionHash(action)
    if (!watchIndex[hash]) {
      yield handleWatchTransactionRequest(action)
    }
  }
}
```

PR #80, opened 2018-03-02
chore: add toLowerCase to all externally obtained addresses

NicoSantangelo:
@cazala I don't like this commit :(
I tried to only add the `lowercase` on external addresses, for example when we're getting it from MetaMask or on the server. Let me know if you have a better idea, or if I missed something

### DIFF
--- a/scripts/initDb.js
+++ b/scripts/initDb.js
@@ -93,7 +93,7 @@ async function normalizeDistrictEntries() {
   for (const districtEntry of districtEntries) {
     await Contribution.insert({
       id: districtEntry.id,
-      address: districtEntry.address,
+      address: districtEntry.address.toLowerCase(),
       district_id: districtEntry.project_id,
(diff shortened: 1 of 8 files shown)


cazala:
@NicoSantangelo we should also change the `seed.js` script to generate addresses in lowercase right? All my mocked publications have mixed upper/lower case addresses, so we should regenerate them after this.

NicoSantangelo:
Yes! I changed that [here](https://github.com/decentraland/marketplace/pull/80/files#diff-01fecf54045496c470be4fd2738ef14c).

You can maybe run this on the server start:

```javascript
const publications = await Publication.find()
for (const pub of publications) {
  await Publication.update({ owner: pub.owner.toLowerCase() }, { tx_hash: pub.tx_hash })
}
```

to update your addresses (disclaimer, wrote it here)

PR #85, opened 2018-03-08
[Feature] View only mode

cazala:
**UPDATED:** Rebased with master!

In this PR:

* The UI handles the wallet errors gracefully and allows the user to browse the website even if they are not connected to Ethereum.

* The wallet connection logic has been abstracted into a separate `Wallet` component instead of having each page do the connection.

* I added a "Sign In" page that will act as a placeholder for now until we make the real one (I need some designs and a copy for that)

* The `/activity`, `/settings` and `/:x/:y/buy` are not accesible via UI but if the user access them by typing the url in the browser it will show them a warning pointing them to the Sign In page.

no-wallet

### DIFF
--- a/webapp/src/Routes.js
+++ b/webapp/src/Routes.js
@@ -16,6 +16,7 @@ import BuyParcelPage from 'components/BuyParcelPage'
 import CancelSalePage from 'components/CancelSalePage'
 import ActivityPage from 'components/ActivityPage'
 import SettingsPage from 'components/SettingsPage'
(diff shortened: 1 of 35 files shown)


NicoSantangelo:
Testing it out but, will need some rebasing :(

NicoSantangelo:
Have a few simple comments will add after the rebase because even tho I only reviewed the new commits, I'm afraid of messing up

cazala:
@NicoSantangelo conflicts fixed, branch rebased!

PR #90, opened 2018-03-14
[Feature] Internationalization

NicoSantangelo:


cazala:
@NicoSantangelo I checked out the branch, npm install'd, and then re-run the server + ui, and I was getting a 404 to `/api/translations/es`. I hardcoded `/api/translations/en` and the request now goes thru, the action `[Success] Fetch Translations` is dispatched, but the placeholders don't get the actual texts, i'm seeing `global.*` all over the place. Am I missing something?

NicoSantangelo:
Hm, I don't understand why you'd get a 404, the server defaults to an empty object when the locale doesn't exist:
image

Nevertheless, I'm testing the entire flow for unknown languages, as mine was english and everything worked out of the box

NicoSantangelo:
@cazala Updated to default to english. You may need to update your localstorage 🙏

PR #92, opened 2018-03-14
[Fix] Publication states and a bunch of minor issues/improvements

cazala:
This PR fixes the following bugs due to state inconsistencies in the redux state:

- I buy X, Y and then I go to the page detail: I’m able to buy it again 
- I publish X, Y on sale and then I go to the detail: I’m able to sell it again
- I cancel the listing on X, Y and then I go to the detail: I’m able to cancel it again

I also let the `Parcel` component re-fetch a parcel even if it is already in the app state, to make sure we always keep the state up to date. This change impacts the following pages:

- ParcelDetailPage
- PublishPage
- BuyParcelPage
- CancelSalePage
- EditParcelPage
- TransferParcelPage
 
Also this PR fixes the following issue:

- Entering `/x/y/cancel-sale` of a cancelled sale still allows cancelation

And also includes the proposal about hiding the sold/cancelled publications in the "On Sale" tab in `ProfilePage`


webapp/src/components/CancelSalePage/CancelSalePage.js
```diff
@@ -44,12 +44,14 @@ export default class CancelSalePage extends React.PureComponent {
                     })}
                   </React.Fragment>
                 ) : (
-                  ''
+                  <React.Fragment>is not for sale.</React.Fragment>
```
NicoSantangelo: We should translate this 🙏

PR #94, opened 2018-03-15
[Feature] Retry button sign in page

NicoSantangelo:
Depends on #90  !!!
will need rebase.

@cazala I tried to stray from adding an `onConnect` method to the `SignIn` page but didn't find a way to "reload" a route. I didn't want to add more redirects either :/
Do you have a better solution?


cazala:
Nice 👌

PR #102, opened 2018-03-16
feat: delete older publications for the same x,y,owner before creating a new one

NicoSantangelo:
@abarmat

### DIFF
--- a/scripts/monitor/persistEvents.js
+++ b/scripts/monitor/persistEvents.js
@@ -7,7 +7,7 @@ import { BlockchainEvent } from '../../src/BlockchainEvent'
 const log = new Log('persistEvents')
 
 export async function persistEvents(lastBlockNumber = null, delay) {
-  delay = delay || env.get('PERSIST_DELAY', 2500)
+  delay = delay || env.get('PERSIST_EVENTS_DELAY', 2500)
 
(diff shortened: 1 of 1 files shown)


scripts/monitor/persistEvents.js
```diff
@@ -53,6 +55,13 @@ export async function processEvent(event) {
       }
       log.info(`[${name}] Creating publication ${contract_id} for ${parcelId}`)
 
+      await Publication.delete({
+        x,
+        y,
+        owner: seller.toLowerCase(),
```
eordano: Do we need this? It may be better to skip `owner` altogether...

abarmat: Do you mean nos filtering by owner?

eordano: Right

NicoSantangelo: I added it to avoid being to eager deleting publications _if_ we're replaying the monitor from the beginning. While the data is not lost, if everything is working correctly publications from other users should be cancelled or sold, no?

abarmat: @eordano do you see a possible problem by constraining to owner?

PR #103, opened 2018-03-16
feat: add terms page, change modal functionality

eordano:


abarmat:
Wouldn't be a good idea to set the links with target blank so we keep the focus on the modal? what do you think?

eordano:
Roger. Updating to new look and feel as well

eordano:
Fixed @abarmat

cazala:
awesome!

PR #105, opened 2018-03-17
fix: `parcel` might be null on parcelUtils#getParcelAttributes

eordano:


cazala:
I'm closing this because since `v0.3.0` parcels always have a `data` prop of type `object`.

PR #106, opened 2018-03-17
fix: Decrease wallet balance on buy. Remove unused method on Publication

NicoSantangelo:


abarmat:
The balance is updated when the tx is sent to the pending state, which is fine! 

The problem is that if the tx fails the balance is not getting updated to the previous value.

NicoSantangelo:
True! updated to change when de tx confirms. Thanks!

cazala:
🥇

PR #109, opened 2018-03-18
fix: intercom position and size

cazala:
This fixes the position and size of the Intercom launcher on the footer. ~The background color needs to be set to `#202433` from the Interdom's admin panel.~ [Done](https://github.com/decentraland/marketplace/pull/109#issuecomment-374010077).

screen shot 2018-03-18 at 12 39 11 pm


abarmat:
Done changing the background in Intercom!

cazala:
@abarmat awesome!

screen shot 2018-03-18 at 12 55 45 pm

PR #122, opened 2018-03-19
feat: added Chinese and Korean

cazala:
Added `zh` and `ko`


NicoSantangelo:
On Chinese

image

cazala:
I'll add the missing translations and update the PR

PR #138, opened 2018-03-20
fix: show 0 balance

cazala:
Before:
screen shot 2018-03-20 at 2 29 21 pm

After:
screen shot 2018-03-20 at 2 26 48 pm

### DIFF
--- a/webapp/src/components/Mana/Mana.js
+++ b/webapp/src/components/Mana/Mana.js
@@ -21,7 +21,7 @@ export default class Mana extends React.PureComponent {
   getLocalizedAmount() {
     const { amount } = this.props
 
-    if (amount) {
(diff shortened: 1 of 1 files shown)


webapp/src/components/Mana/Mana.js
```diff
@@ -21,7 +21,7 @@ export default class Mana extends React.PureComponent {
   getLocalizedAmount() {
     const { amount } = this.props
 
-    if (amount) {
+    if (typeof amount === 'number') {
```
NicoSantangelo: What if we check for `amount > 0` to avoid a possible future bug?

cazala: `>=`, 0 is a valid balance

PR #141, opened 2018-03-20
fix: reset page when sorting

cazala:
This PR makes the page go back to 1 when sorting on the Marketplace page.

Also it improves the contrast of the text on the Genesis Plaza tooltips (black over green instead of white over green).


webapp/src/lib/parcelUtils.js
```diff
@@ -78,7 +78,7 @@ export function getParcelAttributes(id, x, y, wallet, parcels, districts) {
       return {
         label: 'Genesis Plaza',
         description: null,
-        color: 'white',
+        color: 'black',
```
NicoSantangelo: 😱

PR #143, opened 2018-03-20
feat: only store the useful parts of a transaction

NicoSantangelo:
Testing one last strategy

### DIFF
--- a/webapp/src/modules/storage/middleware.js
+++ b/webapp/src/modules/storage/middleware.js
@@ -1,6 +1,7 @@
 import * as storage from 'redux-storage'
 import createStorageEngine from 'redux-storage-engine-localstorage'
 import filter from 'redux-storage-decorator-filter'
+import { env } from 'decentraland-commons'
 
 import {
   FETCH_TRANSACTION_REQUEST,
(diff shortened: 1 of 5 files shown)


NicoSantangelo:
@cazala I don't know how I feel about the second commit. It works, but the code doesn't convince me. I'll point out the flaws I see in comments. If you feel the same way let me know if you can think of another approach. If not, we can simply remove that commit and leave the first onw

webapp/src/modules/storage/selectors.js
```diff
@@ -1,2 +1,2 @@
 export const getState = state => state.storage
-export const isLoading = state => getState(state).loading
+export const isLoading = state => getState(state).loading >= 2
```
NicoSantangelo: I can't tell how many storage middlewares are there without doing some weird state management. That leaves this hardcoded `2` here, which actually means `amount of keys saved`.

I think this is far from optimal, we'll need to remember this next time we're changing the stored state, but I can't figure out a good way to do it

webapp/src/modules/storage/middleware.js
```diff
@@ -13,25 +14,54 @@ import {
...
 import { UPDATE_DERIVATION_PATH } from 'modules/wallet/actions'
 
-export function createStorageMiddleware(storageKey) {
-  const storageEngine = filter(createStorageEngine(storageKey), [
-    'transaction',
-    'translation',
-    ['wallet', 'data', 'locale'],
-    ['wallet', 'data', 'derivationPath']
-  ])
+export function createMainStorageMiddleware() {
```
NicoSantangelo: I decided to create a function for each piece of the store to be stored ( `createMainStorageMiddleware` and `createTransactionStorageMiddleware` ) because using the generic `createStorageMiddleware` felt a bit too much  code for the store, but I wonder if this is too much coupling for this file.

PR #145, opened 2018-03-21
feat: Sanity check (with fix mode)

eordano:


scripts/sanityCheck.js
```diff
@@ -0,0 +1,230 @@
...
+  }
+
+  log.info('[Check 2/2]: Checking parcel ownership from -150,-150 to 150,150')
+  await processParcels(shouldFix)
+}
+
+async function marketplaceCheck(fix) {
+  const marketplace = eth.getContract('Marketplace')
+  const events = await getAllEvents(marketplace)
+  const filterParcels = getParcelIds(events)
```
NicoSantangelo: I find this a bit confusing. `getParcelIds` doesn't return ids but an object, and when I tried to see how it's used in the loop below, the name of the iteration variable is `d` which didn't help me either.

If you agree, maybe something like:

```javascript
const filterParcels = getEventParcels(events)

const tasks = []
for (let { parcelId, lastEvent } of filterParcels) {
  tasks.push(checkParcel(parcelId, lastEvent, marketplace, fix))
}
```

is a bit more descriptive? it's hard to name!

scripts/sanityCheck.js
```diff
@@ -0,0 +1,230 @@
...
+      })
+  })
+}
+
+function getParcelIds(events) {
+  const results = {}
+  for (let event of events) {
+    const str = event.args.assetId
+    if (!str) continue
+    if (event.removed && results[str]) continue
```
NicoSantangelo: I don't understand this condition. Shoudn't that be an `||` ?

scripts/sanityCheck.js
```diff
@@ -0,0 +1,230 @@
...
+  }
+  return Object.values(results)
+}
+
+const NULL =
+  '0x0000000000000000000000000000000000000000000000000000000000000000'
+
+async function checkParcel(parcelId, lastEvent, marketplace, fix) {
+  const res = await decodeAssetId(parcelId)
+  const [x, y] = res.split(',')
```
NicoSantangelo: You can use `Parcel.splitId` if you want here. It does the same thing but abstracts how ids work, just in case

scripts/sanityCheck.js
```diff
@@ -0,0 +1,230 @@
...
+  const publication = await Publication.findInCoordinate(x, y)
+  const auction = await marketplace.instance.auctionByAssetId(parcelId)
```
NicoSantangelo: This `.instance` is not necessary! the contract infers the methods from the ABI :D

scripts/sanityCheck.js
```diff
@@ -0,0 +1,230 @@
...
+    await marketplaceCheck(shouldFix)
+  } catch (err) {
+    console.log(err, err.stack)
+  }
+
+  log.info('[Check 2/2]: Checking parcel ownership from -150,-150 to 150,150')
+  await processParcels(shouldFix)
+}
+
+async function marketplaceCheck(fix) {
```
NicoSantangelo: Can I be annoying and ask for this to be called `shouldFix` everywhere? I really like that name from `sanityCheck` because it's clear that it's a booelan value

scripts/sanityCheck.js
```diff
@@ -0,0 +1,230 @@
...
+      return
+    }
+  }
+}
+
+function deletePublication(x, y) {
+  return Publication.delete({
+    x,
+    y
+  })
```
NicoSantangelo: This scares me. This is potentially removing publications from older owners which might be `cancelled` or `sold` which tampers with our ability to show it on the UI. The method is not really `deletePublication` but `deletePublications`.

Maybe we can use a timestamp or the owner? Do you think this is a problem?

NicoSantangelo:
One last thing! I tried to test it on ropsten and I think the contract might not be updated because I get:

```
-32 -143 different id in db 0x72b69b1e6c61f329cb9bef785c867b816953bd9c3e608f8d42577a684fa4b316 vs in blockchain 0x
-39 -146 different id in db 0x3072f711c078be9d2040968900a9843acc5848760d0c0529fe46451c09785682 vs in blockchain 0x
-39 -144 different id in db 0x2eac3c891e1d4404629e537f3fc4c3e3bfc4e0e603fa0e46907fd21df242efcc vs in blockchain 0x
-31 -146 different id in db 0x84b8334c99a8f5d70563562e1bf4c6acfbec1941689372e648a5cdc5a1266bc5 vs in blockchain 0x
-32 -147 different id in db 0xbc3b6248aa93c8a713eefae5e0b1fc4ae02cb8e4321f0082e2ddf1e3a081fc6b vs in blockchain 0x
```

if that's the case it's fine, but I wanted to make sure the empty address looked like `0x0000000000000000000000000000000000000000000000000000000000000000` and not like `0x`

scripts/sanityCheck.js
```diff
@@ -0,0 +1,230 @@
...
+}
+
+function getAllEvents(marketplace) {
+  return new Promise((resolve, reject) => {
+    marketplace.instance
+      .allEvents({
+        fromBlock: 0,
+        toBlock: 'latest'
+      })
+      .get(function(error, events) {
```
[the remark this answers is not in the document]
eordano: Yes I was aware! Had a problem trying to set it up :( It looked like my callback was never called

NicoSantangelo:
@eordano We had a few problems deploying the contract in ropsten and making it work. It's working perfectly but I have one last question:

- Do we want this check? https://github.com/decentraland/marketplace/pull/145/files#diff-df5235e139d74c901c00f864aff706feR183
I get this filters un-owned parcels but what if they get an owner later on and for whatever reason the indexer doesn't pick it up, the sanity-check will ignore that scenario.
What do you think? do we want to change that?

eordano:
@NicoSantangelo Addressed! I changed it to pick up the owner from the contract and verify myself against that (`isOwner` was not enough)

NicoSantangelo:
Testing !

PR #147, opened 2018-03-21
feat: remove "most expensive" sort

cazala:
so folks don't abuse it

### DIFF
--- a/src/Publication/PublicationRequestFilters.js
+++ b/src/Publication/PublicationRequestFilters.js
@@ -8,8 +8,9 @@ const DEFAULT_VALUES = Object.freeze({
 })
 
 const ALLOWED_VALUES = Object.freeze({
-  sort_by: ['price', 'expires_at', 'created_at'],
-  sort_order: ['ASC', 'DESC']
+  price: ['ASC'],
+  created_at: ['DESC'],
+  expires_at: ['ASC']
 })
 
 export class PublicationRequestFilters {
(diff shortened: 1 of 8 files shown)


NicoSantangelo:
We should remove it from the server too, to avoid URL abuse

src/Publication/PublicationRequestFilters.js
```diff
@@ -26,53 +27,27 @@ export class PublicationRequestFilters {
   }
 
   sanitize(req) {
+    let by = server.extractFromReq(this.req, 'sort_by')
```
NicoSantangelo: `server.extractFromReq` throws if the key is not present.
Maybe we should wrap this in a `trycatch`? I'll leave it up to you, I think it might not be a bad idea to throw here

src/Publication/PublicationRequestFilters.js
```diff
@@ -26,53 +27,27 @@ export class PublicationRequestFilters {
...
+    let order = server.extractFromReq(this.req, 'sort_order')
+    let limit = server.extractFromReq(this.req, 'limit')
+    let offset = server.extractFromReq(this.req, 'offset')
+
+    by = by in ALLOWED_VALUES ? by : DEFAULT_VALUES['sort_by']
+    order = ALLOWED_VALUES[by].includes(order.toUpperCase())
+      ? order
+      : ALLOWED_VALUES[by][0]
+    limit = Math.max(Math.min(100, limit), 0)
+    offset = Math.max(offset, 0)
```
NicoSantangelo: Maybe this 0 and the above 100 should be obtain from `DEFAULT_VALUES` ? or just delete `DEFAULT_VALUES` completely, no?

cazala: I removed DEFAULT_VALUES

src/Publication/PublicationRequestFilters.js
cazala: I removed DEFAULT_VALUES so it will throw (the UI doesn't hit that endpoint without all the params tho)

PR #164, opened 2018-03-25
feat: add command line helper tool for encoding decoding assetId

abarmat:
Usage:

`mktcli.js encode '(1,2)'`
`mktcli.js decode 3.40282366920938463463374607431768211457e+38`

### DIFF
--- a/scripts/mktcli.js
+++ b/scripts/mktcli.js
@@ -3,6 +3,7 @@
 import { eth, Contract, Log, cli, contracts } from 'decentraland-commons'
 import { db } from '../src/database'
 import { Parcel } from '../src/Parcel'
+import { Publication } from '../src/Publication'
 import { loadEnv, parseCLICoords } from './utils'
 import { decodeAssetId, encodeAssetId } from './monitor/utils'
 
@@ -23,6 +24,7 @@ const main = {
         } catch (err) {
           logError(err)
         }
+        process.exit()
       })
 
     program
@@ -43,6 +45,7 @@ const main = {
         } catch (err) {
           logError(err)
         }
+        process.exit()
       })
 
     program
@@ -65,6 +68,7 @@ const main = {
         } catch (err) {
           logError(err)
         }
+        process.exit()
       })
 
     program
@@ -81,6 +85,7 @@ const main = {
         } catch (err) {
           logError(err)
         }
+        process.exit()
       })
 
     program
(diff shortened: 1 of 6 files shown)


scripts/mktcli.js
```diff
@@ -0,0 +1,105 @@
...
+
+      try {
+        const coords = await decodeAssetId(options)
+        log.info(`(decode) str:${options} => coords:(${coords})`)
+      } catch (err) {
+        logError(err)
+      }
+    })
+
+    program.command('encode').action(async options => {
```
NicoSantangelo: _This goes for each command that receives a parameter_
I think we should add each required parameter as an explicit parameter of the command like we do here: https://github.com/decentraland/marketplace/blob/master/scripts/seed.js#L17 so, for example the help command works correctly:

image

scripts/mktcli.js
```diff
@@ -0,0 +1,105 @@
...
+
+const logError = err => log.error('ERR: ' + err.message)
+const parseCoords = coords =>
+  coords
+    .replace('(', '')
+    .replace(')', '')
+    .split(',')
+
+const landRegistry = () => eth.getContract('LANDRegistry')
+const marketplace = () => eth.getContract('Marketplace')
```
NicoSantangelo: Maybe this should be called `getMarketplace` or  `getMarketplaceContract` ? so we know that they're returning something

scripts/mktcli.js
```diff
@@ -0,0 +1,105 @@
+#!/usr/bin/env babel-node
+
+import { eth, Log, cli, contracts } from 'decentraland-commons'
+import { loadEnv } from './utils'
+import { decodeAssetId, encodeAssetId } from './monitor/utils'
+
+const log = new Log('mktcli')
```
NicoSantangelo: If we're having a _global_ cli, it seems the perfect place to put the rest of the scripts we're using independently. So for example we could do:

```bash
npm run cli monitor
npm run cli renew-blockchain
# etc
```

because if we don't we risk continuing to add more and more standalone scripts.

This can be a separate PR!, but what do you think? the alternative is to rename this file to something along the lines of `land-manager` maybe?

scripts/mktcli.js
```diff
@@ -0,0 +1,105 @@
...
+const parseCoords = coords =>
+  coords
+    .replace('(', '')
+    .replace(')', '')
+    .split(',')
+
+const landRegistry = () => eth.getContract('LANDRegistry')
+const marketplace = () => eth.getContract('Marketplace')
+
+const main = {
```
NicoSantangelo: The commands don't work with negative coords

scripts/mktcli.js
```diff
@@ -0,0 +1,105 @@
...
+      }
+    })
+
+    program.command('land-owner').action(async options => {
+      if (!options.length) return
+
+      try {
+        const [x, y] = parseCoords(options)
+        const contract = landRegistry()
+        const owner = await contract.ownerOfLand(x, y)
```
NicoSantangelo: maybe we could add an `if Contract.isEmptyAddress() => search if it's a district` ?

PR #165, opened 2018-03-25
fix: add handling of events passed as array

eordano:
Updating the dependencies for `commons`, it seems that `eventData` might be an array of events. This change prevents this.

Also: Fix `monitor` as it wasn't connecting to `RPC_URL` for me

### DIFF
--- a/scripts/monitor/handlers/index.js
+++ b/scripts/monitor/handlers/index.js
@@ -4,6 +4,9 @@ import { BlockchainEvent } from '../../../src/BlockchainEvent'
 const log = new Log('handlers')
 
 export async function store(eventData) {
+  if (eventData.length) {
+    return Promise.all(eventData.map(store))
+  }
   if (eventData.removed) return
 
   const { event, transactionHash, blockNumber, logIndex } = eventData
--- a/scripts/monitor/index.js
+++ b/scripts/monitor/index.js
@@ -1,6 +1,6 @@
 #!/usr/bin/env babel-node
 
-import { eth, Log } from 'decentraland-commons'
+import { env, eth, contracts, Log } from 'decentraland-commons'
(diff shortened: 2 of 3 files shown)


scripts/monitor/handlers/index.js
```diff
@@ -4,6 +4,9 @@ import { BlockchainEvent } from '../../../src/BlockchainEvent'
 const log = new Log('handlers')
 
 export async function store(eventData) {
+  if (eventData.length) {
```
NicoSantangelo: huh, wasn't aware this might happen. The docs say that the callback receives a result `object` https://github.com/ethereum/wiki/wiki/JavaScript-API#callback-return

That said I think is the responsibility of the StoreCli to report each event to the handler, meaning changing this line instead:

https://github.com/decentraland/marketplace/blob/master/scripts/monitor/StoreCli.js#L39

from:

```javascript
await handler(eventData)
// to something like 
if (eventData.length) await Promise.all(eventData.map(handler)
 else await handler(eventData)
```

what do you think?

scripts/monitor/index.js
```diff
@@ -17,7 +17,10 @@ Promise.resolve()
   })
   .then(() => {
     log.debug('Connecting to Ethereum node')
-    return eth.connect()
+    return eth.connect({
```
NicoSantangelo: eth (or commons for that matter) doesn't relay on the ENV so if you want to use the RPC_URL you have to explicity use it like you did here, that's why it wasn't working.

I think the contracts can be left out because the EventMonitor sets them dynamically here https://github.com/decentraland/marketplace/blob/master/scripts/monitor/EventMonitor.js#L67

NicoSantangelo:
Feature already present in master

PR #176, opened 2018-03-30
feat: redesign

cazala:
This PR applies the following design changes:

Navbar:

- Added background to loading animation
- Added support for "go back" mode (for modals and settings page)

Profile:

- Added profile Blockie (only for other users' profiles, it's not visible on My Land)

Marketplace: 

- Added number of parcels on sale in the tab

Settings:

- Two-column design with user Blockie

Edit:
Transfer:
Publish:
CancelSale:
Buy:

- Two column design with parcel preview, refactored those screens to use a `ParcelModal` component that contains the layout

### DIFF
--- a/webapp/src/components/CancelSalePage/CancelSalePage.js
+++ b/webapp/src/components/CancelSalePage/CancelSalePage.js
@@ -1,12 +1,11 @@
 import React from 'react'
 import PropTypes from 'prop-types'
-
-import { Container, Header, Grid, Button } from 'semantic-ui-react'
-import ParcelName from 'components/ParcelName'
+import { Link } from 'react-router-dom'
 import Parcel from 'components/Parcel'
-
-import { formatMana } from 'lib/utils'
-import { t } from 'modules/translation/utils'
+import ParcelModal from 'components/ParcelModal'
+import { buildCoordinate } from 'lib/utils'
+import { t, t_html } from 'modules/translation/utils'
+import { locations } from 'locations'
 
 import './CancelSalePage.css'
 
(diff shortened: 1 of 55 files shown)


webapp/src/components/CancelSalePage/CancelSalePage.js
```diff
@@ -24,53 +23,42 @@ export default class CancelSalePage extends React.PureComponent {
...
+    )
+    return (
+      <React.Fragment>
+        {publication ? (
+          <React.Fragment>
+            {t_html('parcel_cancel.about_to_cancel', { parcel })}
+          </React.Fragment>
+        ) : (
+          <React.Fragment>
+            {t_html('parcel_cancel.not_for_sale', { parcel })}
```
NicoSantangelo: I think the `Fragment` isn't necessary here (on the `t_html` calls)

webapp/src/components/ParcelModal/ParcelModal.js
```diff
@@ -0,0 +1,83 @@
...
+            <span className="modal-subtitle">{subtitle}</span>
+          </div>
+          <br />
+          {children ? (
+            <React.Fragment>
+              <div className="modal-children">{children}</div>
+              <br />
+            </React.Fragment>
+          ) : null}
+          {hasCustomFooter ? null : (
```
NicoSantangelo: Just a suggestion, when reading the prop from the implementation on other components, I thought the custom footer was the children prop. Might be just me, but maybe this prop could be `hasFooterButtons` or something like that?

webapp/src/modules/location/selectors.js
```diff
@@ -7,3 +7,21 @@ export const getParams = state => getState(state).params
...
+  if (state.router && state.router.location.pathname) {
+    const lastPartOfUrl = state.router.location.pathname.split('/').pop()
+    switch (lastPartOfUrl) {
+      case 'edit':
+      case 'buy':
+      case 'sell':
+      case 'cancel-sale':
+      case 'transfer':
+      case 'settings':
+        return true
```
NicoSantangelo: I might be crazy, but can we try to use [matchParam](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/matchPath.md) with our `locations` object. What do you think?

webapp/src/components/CancelSalePage/CancelSalePage.js
```diff
@@ -28,19 +28,9 @@ export default class CancelSalePage extends React.PureComponent {
...
-          <React.Fragment>
-            {t_html('parcel_cancel.about_to_cancel', { parcel })}
-          </React.Fragment>
-        ) : (
-          <React.Fragment>
-            {t_html('parcel_cancel.not_for_sale', { parcel })}
-          </React.Fragment>
-        )}
-      </React.Fragment>
-    )
```
NicoSantangelo: ❤️

PR #183, opened 2018-04-03
feat: add endpoint to get asset related info

abarmat:
1) Some debatable things are: should it be part of the /parcels resources or not? As the query was not by (x,y) and the return is a bit more generic than exactly a parcel I created /assets, give me your opinion about this, please.

2) Test a parcel with no metadata: `http://localhost:5000/api/assets/0`
3) Test a parcel with metadata: `http://localhost:5000/api/assets/0xffffffffffffffffffffffffffffffc90000000000000000000000000000007b`

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -215,7 +215,7 @@ export async function getAssetId(req) {
   const parcel = await Parcel.findOne({ asset_id: strAssetId })
 
   if (!parcel) {
-    throw new Error(`Could not find a valid asset for the ID "${strAssetId}"`)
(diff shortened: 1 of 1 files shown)


abarmat:
Sample output format:

```{"ok":true,"data":{"id":"-55,123","owner":"0xd37398e2e24a12abcdb9966c199b5cd562c3b305","created_at":"2018-03-15T14:46:59.196Z","updated_at":"2018-04-02T14:58:38.906Z","description":"","name":"exchange","img_url":"/api/parcels/-55/123/preview"}}```

abarmat:
Dropped this idea, will use the same interface for our client and others.

PR #198, opened 2018-04-21
feat: better display of MANA and ETH rate

eordano:


abarmat:
Do we need the MANA to ETH exchange rate? Most of the time it is 0 as it's a very small number.

screen shot 2018-04-30 at 11 49 14

eordano:
You're right... let's keep the X MANA = 1 ETH one

PR #200, opened 2018-04-25
feat: new parcel card

NicoSantangelo:
Does a lot more than just redisigning the parcel card:

- Updates the entire API to be a bit more consistent and return parcels with publications when possible
- Updates the entire redux saga/store/selectors to use parcels with publications
- Extracts <Tag /> into it's own component

Steps:
- `npm run migrate up`
- `npm run tag`
- `psql $CONNECTION_STRING -f scripts/addLastTransferedAt.sql`


webapp/src/components/ParcelCard/ParcelCard.js
```diff
@@ -0,0 +1,70 @@
...
+                <Card.Meta title={formatDate(publication.expires_at)}>
+                  <PublicationExpiration publication={publication} />
+                </Card.Meta>
+                <div className="mana">
+                  <Mana amount={parseFloat(publication.price, 10)} />
+                </div>
+              </React.Fragment>
+            ) : (
+              <Card.Meta>
+                Acquired at&nbsp;
```
cazala: missing translation

PR #201, opened 2018-04-25
chore: fix env var warning

abarmat:


NicoSantangelo:
Good catch! but maybe we can copy the implementation of the other scripts? so we remain consistent.

Some examples:
- https://github.com/decentraland/marketplace/blob/master/scripts/addAuctionOwners.js#L11
- https://github.com/decentraland/marketplace/blob/master/scripts/addBlockTimes.js#L12

what do you think?

abarmat:
For sure, will change it!

PR #211, opened 2018-05-10
Update stats counter to include more interactions and contributors

abarmat:


src/Dashboard/Dashboard.service.js
```diff
@@ -34,16 +38,29 @@ export class DashboardService {
...
-      `SELECT COUNT(DISTINCT(owner)) as count
-        FROM ${this.Publication.tableName}`
+      `SELECT COUNT(DISTINCT(A.address)) 
+        FROM (
+          SELECT owner as address 
+            FROM ${this.Publication.tableName} 
+          UNION 
+          SELECT args->>'from' AS address 
+            FROM ${this.BlockchainEvent.tableName} 
+            WHERE name IN ('Transfer', 'Update')
```
NicoSantangelo: Can we use `BlockchainEvent` `EVENTS` here?

something like:

```javascript
  async countActiveUsers() {
    const { parcelTransfer, parcelUpdate } = this.BlockchainEvent.EVENTS

    return await this.count(
      `SELECT COUNT(DISTINCT(A.address))
        FROM (
          SELECT owner as address
            FROM ${this.Publication.tableName}
          UNION
          SELECT args->>'from' AS address
            FROM ${this.BlockchainEvent.tableName}
            WHERE name IN ($1, $2)
        ) AS A`,
      [parcelTransfer, parcelUpdate]
    )
  }
```

PR #219, opened 2018-05-15
feat: estates backend

NicoSantangelo:


webapp/src/modules/publication/selectors.js
```diff
@@ -34,3 +37,18 @@ export const getPublications = createSelector(
...
+
+  if (parcel) {
+    const publications = getData(state)
+    const parcelPublications = findAssetPublications(
+      publications,
+      parcel,
+      PUBLICATION_STATUS.open
+    )
+    return parcelPublications[0]
+  }
```
cazala: could we return `null` instead of `undefined` if the parcel is not in the state?

migrations/1526294607563_parcels-add-estate-data.js
```diff
@@ -0,0 +1,9 @@
+import { Parcel } from '../src/Parcel'
+
+exports.up = pgm => {
+  const tableName = Parcel.tableName
+
+  pgm.addColumns(tableName, {
+    in_estate: { type: 'BOOLEAN', default: false, allowNull: false }
```
cazala: I think we will need the actual `estate_id` instead of a boolean here, so we can blur the lines between parcels of the same estate on the map (same way we do with districts)

webapp/src/modules/publication/selectors.js
NicoSantangelo: 👍 !

NicoSantangelo:
Updated!

cazala:
🥇

PR #221, opened 2018-05-15
feat: manage parcel

cazala:
Included in this PR:

+ Give permission to update the contents of a parcel
+ Revoke/Change permissions already set

Here an example of giving permission to an address, followed by an example of how to revoke that permission:

manage parcel


NicoSantangelo:
image

Couple of things:
- It's the only "modal like" page that has the navbar. Is that intentional?
- I think its a bit weird to have the `cancel` and `submit` buttons when the two actions are `change/revoke`. It would make more sense to hide them if the we have the back button on the top left. What do you think?
This also gets exacerbated by the fact that canceling a revoke takes me to `back` to the parcel detail page, which is a bit surprising when I come from the `change/revoke` menu.
- I love the realtime blockie

-------




Mobile blockie, also, from 768px to ~600px it looks like this:

image

which I don't know if its intentional

-------

image

I got this prop-type error but I don't really know how. If you can't reproduce it I can try giving you more info of my current state

--------




The text on the sell button "falls down" but not on transfer.

--------

Should we add an error message for invalid addresses? I think it might be clear enough as it is, but it can be confusing for people if they copied the wrong address, they think it's okay and all they see it's a red underline on the input.
Not sure about this one

webapp/src/lib/parcelUtils.js
```diff
@@ -147,3 +148,20 @@ export function getParcelAttributes(id, x, y, wallet, parcels, districts) {
...
+    if (
+      eth.utils.isValidAddress(address) &&
+      !Contract.isEmptyAddress(address)
+    ) {
+      return address
+    }
+    return null
+  } catch (error) {
+    return null
+  }
```
NicoSantangelo: maybe we can keep only one `return null` after the catch?

webapp/src/modules/parcels/sagas.js
```diff
@@ -85,3 +89,27 @@ function* handleEditParcelsRequest(action) {
     yield put(editParcelFailure(parcel, error.message))
   }
 }
+
+function* handleManageParcelsRequest(action) {
+  try {
+    const { parcel, address, revoked } = action
+    const { x, y } = parcel
```
NicoSantangelo: _nitpick_: we can move this line above the `try` to avoid repeating it on the catch

webapp/src/modules/parcels/utils.js
```diff
@@ -1,5 +1,13 @@
 import { buildCoordinate } from 'lib/utils'
 
+export function shouldConnect(parcelA, parcelB, prop) {
```
NicoSantangelo: _Future work_: we may want to add handleParcelRequest to the webworker? to avoid having to repeat this logic here

webapp/src/components/ParcelPreview/ParcelCanvas/ParcelCanvas.js
```diff
@@ -129,25 +129,26 @@ export default class ParcelPreview extends React.PureComponent {
...
+    //     center: {
+    //       x: nextProps.x,
+    //       y: nextProps.y
+    //     },
+    //     pan: {
+    //       x: 0,
+    //       y: 0
+    //     }
+    //   }
+    // }
```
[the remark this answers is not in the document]
cazala: This disables recentring when changing the x/y props (controlled) but it looks laggy. In a future PR I would change the x/y props to initialX/initialY (so if you change it afterwards it doesn't recenter) and then let x/y do the recentering if changed from outside, but currently it makes the parcel detail look weird when you select a new parcel (it forces the map to recenter on that parcel but it feels too violent)

webapp/src/modules/parcels/utils.js
cazala: +1

webapp/src/components/ManageParcelPage/ManageParcelForm/ManageParcelForm.js
```diff
@@ -0,0 +1,184 @@
...
+          </Button>
+
+          <Button
+            type="submit"
+            primary={true}
+            disabled={
+              this.isEmptyAddress() ||
+              isTxIdle ||
+              loading ||
+              !dirty ||
```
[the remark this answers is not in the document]
cazala: `dirty` is used cos if you click on `change` you can hit submit until you actually change the address (isValidAddress returns true cos the previous address was valid)

cazala:
☝️ all the mobile ui issues have been addressed and I added an `undo` button for the revoke action to make it clear that the user doesn't have to hit cancel

PR #222, opened 2018-05-16
feat: mortgage request

nachomazzara:
- [x] Approve Mortgage to use RCN tokens
- [x] Approve Mortgage to use Mana Tokens
- [x] Checkboxes at Settings page
- [x] Transactions at Activity page
- [x] Create Mortgage Request
- [x] See Mortgage Status at Parcel Detail
- [x] See Mortgages
- [x] Translations
- [x] Tests

## NOTES 
Set `react` and `react-dom` to 16.2.0 in order to avoid version updates until we migrate


src/Parcel/Parcel.model.js
```diff
@@ -96,7 +99,9 @@ export class Parcel extends Model {
     return await this.db.query(
       SQL`SELECT *, (
         ${PublicationQueries.findLastParcelPublicationJsonSql()}
-      ) as publication
+      ) as publication, ARRAY(
+        ${MortgageQueries.findParcelMortgageJsonSql()}
+      ) as mortgages
```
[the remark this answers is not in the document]
nachomazzara: Totaly agree 🙂

.eslintignore
```diff
@@ -0,0 +1,2 @@
+# /node_modules/* and /bower_components/* in the project root are ignored by default
+webapp
```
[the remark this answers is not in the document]
nachomazzara: I added it because We were linting `webapp/node_modules`. 

If we add it to the script. We need to lint spec, scripts, etc. too not only src.

Using an ignore file give us the possibility to exclude whatever we want.

What do you think?

migrations/1525973461155_mortgage-create.js
```diff
@@ -0,0 +1,61 @@
...
+        notNull: true
+      },
+      status: {
+        type: 'TEXT',
+        notNull: true,
+        default: Mortgage.STATUS.open,
+        check: `status IN (${mortgageStatus})`
+      },
+      x: { type: 'INT', notNull: true },
+      y: { type: 'INT', notNull: true },
```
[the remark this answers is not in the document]
nachomazzara: Makes sense. 

To answer your question: Buy states by a mortgage is not going to happen in a short/middle term. 

We need to test mortgaged parcel first, see kpis, etc.

scripts/monitor/processEvents.js
```diff
@@ -165,12 +209,55 @@ export async function processEvent(event) {
...
+        await rcnEngineContract.getExpirationRequest(
+          eth.utils.toBigNumber(loanId)
+        )
+      ])
+
+      const exists = await Mortgage.count({ tx_hash })
+      if (exists) {
+        log.info(`[${name}] Mortgage ${tx_hash} already exists`)
+        return
+      }
```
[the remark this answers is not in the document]
nachomazzara: Oops 🤭

scripts/renewBlockchainData.js
```diff
@@ -50,7 +49,7 @@ export async function updateParcelsData(parcels) {
       )
     },
     batchSize: BATCH_SIZE,
-    retryAttempts: 20
+    retryAttempts: 200
```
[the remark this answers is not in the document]
nachomazzara: Sometimes with a batch of 100 i have more than 20 retries.

Also if the geth node goes down, you will have time to set it up again.

Anyways, we still prefer to leave it as 20 I’m ok

src/Parcel/Parcel.model.js
```diff
@@ -161,4 +166,13 @@ export class Parcel extends Model {
   isRoad() {
     return District.isRoad(this.attributes.district_id)
   }
+
+  static async parcelsMortgagesByBorrower(borrower) {
```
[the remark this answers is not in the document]
nachomazzara: Ok, sorry for that 👍

webapp/src/components/Parcel/Parcel.js
```diff
@@ -19,6 +21,8 @@ export default class Parcel extends React.PureComponent {
     isConnecting: false,
     isLoading: false,
     ownerOnly: false,
+    notOwnerOnly: false,
+    withPublications: false,
```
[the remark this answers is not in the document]
nachomazzara: I use it to redirect the user if it is the owner, eg you shuold not request a mortgage for your own parcels.

`notOnlyOwner` = all users except owner.

Maybe I should find a better name

`ownerNotAllowed`, `notOwner`?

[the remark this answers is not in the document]
nachomazzara: I use it to redirect the user if it is the owner, eg you shuold not request a mortgage for your own parcels.

`notOnlyOwner` = all users except owner.

Maybe I should find a better name

`ownerNotAllowed`, `notOwner`?

webapp/src/components/ParcelDetailPage/ParcelDetailPage.js
```diff
@@ -54,7 +56,8 @@ export default class ParcelDetailPage extends React.PureComponent {
       districts,
       publications,
       onBuy,
-      onParcelClick
+      onParcelClick,
+      user
```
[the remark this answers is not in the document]
nachomazzara: I name it `user` as the current user seeing the Daap because as it is the `parcelDetailPage`, the user is not always the `borrower` (only for mortgages)

webapp/src/lib/utils.js
```diff
@@ -71,3 +71,12 @@ export function distanceInWordsToNow(date) {
     locale: getCurrentLocale()
   })
 }
+
+/**
+ * Returns if a date is expired
+ * @param {number} - expires_at
+ * @returns {boolean}
+ */
+export function isExpired(expires_at) {
```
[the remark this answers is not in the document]
nachomazzara: Yes! That was the first thing I made and forgot to refactor! 

Good catch!

webapp/src/modules/mortgage/utils.js
```diff
@@ -0,0 +1,44 @@
...
+  return Math.ceil(days) * 24 * 60 * 60
+}
+
+export function getLoanMetadata() {
+  const mortgageManagerContract = eth.getContract('MortgageManager')
+  return `#mortgage #required-cosigner:${mortgageManagerContract.address}`
+}
+
+export function isOpen(publication) {
+  return hasStatus(publication, MORTGAGES_STATUS.open)
```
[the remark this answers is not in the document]
nachomazzara: idem, refactor needed

webapp/src/modules/mortgage/utils.js
```diff
@@ -0,0 +1,44 @@
...
+    !isExpired(mortgage)
+  )
+}
+
+export function getOpenMortgagesByBorrower(mortgages, borrower) {
+  return mortgages.filter(
+    mortgage =>
+      mortgage.borrower === borrower &&
+      mortgage.status !== MORTGAGES_STATUS.cancelled
+  )
```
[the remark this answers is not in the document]
nachomazzara: idem

nachomazzara:
Thanks, @NicoSantangelo  for the PR!!! it was helpful.

Related to tracking with Segment, I will do it as a feature un another branch (PR) same with the feature of hiding mortgage if env is not development

nachomazzara:
Remove `webapp` from root lint because for some reason it breaks with `react@16.3.0` deprecations

src/Mortgage/Mortgage.model.js
```diff
@@ -0,0 +1,48 @@
...
+          AND status != '${raw(this.STATUS.cancelled)}'
+          ORDER BY created_at DESC`
+    )
+  }
+
+  static findActivesInCoordinate(assetId) {
+    return this.db.query(
+      SQL`SELECT * FROM ${raw(this.tableName)}
+        WHERE asset_id = ${assetId}
+          AND status != '${raw(this.STATUS.cancelled)}'
```
NicoSantangelo: I think this can be just `${this.STATUS.cancelled}`instead of  `'${raw(this.STATUS.cancelled)}'`

src/Mortgage/Mortgage.router.js
```diff
@@ -0,0 +1,67 @@
...
+  async getActiveMortgagesInCoordinate(req) {
+    const x = server.extractFromReq(req, 'x')
+    const y = server.extractFromReq(req, 'y')
+    const status = server.extractFromReq(req, 'status')
+
+    let mortgages = []
+    if (status === 'active') {
+      mortgages = await Mortgage.findActivesInCoordinate(Parcel.buildId(x, y))
+    }
+    return mortgages
```
NicoSantangelo: this is a bit weird to me, if we're only supporting active, why does the endpoint have an `status` param?

src/Translation/locales/fr.json
```diff
@@ -68,18 +68,25 @@
     "connect": "Connection",
     "connecting": "Connexion",
     "contributions": "Contributions",
+    "expired_at": "Temps expiré}",
```
NicoSantangelo: Whoops careful with the `}` here

src/Translation/locales/es.json
```diff
@@ -67,17 +67,24 @@
     "connect": "Conexión",
     "connecting": "Conectando",
     "contributions": "Contribuciones",
+    "expired_at": "Tiempo expirado}",
```
NicoSantangelo: Whoops careful with the } here

src/Translation/locales/es.json
```diff
@@ -253,10 +303,18 @@
...
+    "create_mortgage":
+      "Solicitó una hipoteca para comprar el paquete {parcel_link}",
     "disapproved":
       "Has desautorizado al {marketplace_contract_link} para operar MANA por tu cuenta.",
     "edit":
       "Has editado {parcel_link} con el nombre \"{name}\" y la descripción \"{description}\"",
+    "mortgage_mana":
+      "Usted {acción} el {mortgage_contract_link} para operar MANA en su nombre.",
+    "mortgage_rcn":
+      "Usted {acción} el {mortgage_contract_link} para operar RCN en su nombre.",
```
NicoSantangelo: Whoops careful with the `{acción}` here

src/Translation/locales/es.json
```diff
@@ -105,6 +112,38 @@
...
+    "duration": "Duración",
+    "duration_placeholder": "Duración en días",
+    "errors": {
+      "amount_minimum": "La cantidad debe ser mayor que {cantidad}",
+      "duration_gte_payable":
+        "La duración debe ser mayor o igual que la paga en días",
+      "maximum_interest_rate": "La tasa debe ser inferior a {máximo}",
+      "minimum_duration": "La duración debe ser mayor a {mínimo} días",
+      "minimum_interest_rate": "La tasa debe ser mayor que {mínimo}",
+      "minimum_payable": "Pagable a debe ser mayor a {mínimo} días",
```
NicoSantangelo: Careful with the keys inside the `{}`

src/Translation/locales/fr.json
```diff
@@ -257,10 +307,18 @@
...
+    "create_mortgage":
+      "Vous avez demandé une hypothèque pour acheter le colis {parcel_link}",
     "disapproved":
       "Vous avez autorisé le {marketplace_contract_link} à transférer des fonds en votre nom",
     "edit":
       "Vous avez mis à jour {parcel_link} avec le nom \"{name}\" et la description \"{description}\"",
+    "mortgage_mana":
+      "Vous {actionnez} sur le {mortgage_contract_link} pour exploiter MANA en votre nom.",
+    "mortgage_rcn":
+      "Vous {actionnez} sur le {mortgage_contract_link} pour exploiter RCN en votre nom.",
```
NicoSantangelo: careful with `{actionnez}` here

src/Translation/locales/ko.json
```diff
@@ -246,10 +295,18 @@
...
+    "create_mortgage":
+      "귀하는 저당 {parcel_link}을 구입하라고 모기지에 요청했습니다.",
     "disapproved":
       "귀하를 대신하여 {marketplace_contract_link}의 MANA 운영을 허가받지 않았습니다.",
     "edit":
       "{parcel_link}을(를) \"{name}\" 이름 및 설명 \"{description}\" (으)로 수정했습니다.",
+    "mortgage_mana":
+      "귀하를 대신하여 MANA를 운영하기위한 {행동} {mortgage_contract_link}",
+    "mortgage_rcn":
+      "귀하를 대신하여 RCN을 운영하기위한 {행동} {mortgage_contract_link}",
```
NicoSantangelo: carful with `{행동}` here

src/Translation/locales/zh.json
```diff
@@ -105,6 +112,35 @@
...
+    "duration": "持续时间",
+    "duration_placeholder": "持续时间以天计",
+    "errors": {
+      "amount_minimum": "金额应该大于{金额}",
+      "duration_gte_payable": "持续时间应该大于或等于在天付款",
+      "maximum_interest_rate": "费率应该低于{最大值}",
+      "minimum_duration": "持续时间应该大于{最短}天",
+      "minimum_interest_rate": "费率应该大于{最小值}",
+      "minimum_payable": "应付时间应该大于{最短}天",
+      "parcel_price": "所需金额应低于包裹价格"
```
NicoSantangelo: same as spanish, careful with the keys

webapp/src/components/BuyParcelByMortgagePage/BuyParcelByMortgagePage.js
```diff
@@ -0,0 +1,108 @@
...
+                    </Link>
+                  ),
+                  parcel_price: formatMana(publication.price)
+                })}
+                hasCustomFooter
+              >
+                <MortgageForm
+                  parcel={parcel}
+                  publication={publication}
+                  isTxIdle={false}
```
NicoSantangelo: Is this intended to change based on some data?
Same question for `isDisabled`

webapp/src/components/BuyParcelByMortgagePage/MortgageForm/MortgageForm.js
```diff
@@ -0,0 +1,324 @@
...
+          <Grid.Row columns={3}>
+            <Grid.Column mobile={16} computer={5}>
+              <Form.Field>
+                <label>{t('mortgage.amount')}</label>
+                <Input
+                  type="number"
+                  placeholder={t('mortgage.amount_placeholder')}
+                  value={amount}
+                  required={true}
+                  onChange={e => this.handleChangeNumber(e, 'amount')}
```
NicoSantangelo: I like this, but I'm unsure if it's worth it performance wise. @cazala what do you think?

webapp/src/components/MortgageActions/MortgageActions.container.js
```diff
@@ -0,0 +1,21 @@
...
+    mortgage
+  }
+}
+
+const mapDispatch = (dispatch, { mortgage }) => ({
+  onCancel: () =>
+    dispatch(
+      cancelMortgageRequest(mortgage.mortgage_id, {
+        assetId: mortgage.asset_id
+      })
```
NicoSantangelo: I think the signature for cancelMortgageRequest is `cancelMortgageRequest(mortgageId, assetId)` so it should be `cancelMortgageRequest(mortgage.mortgage_id, mortgage.asset_id)`

webapp/src/lib/utils.js
```diff
@@ -71,3 +72,37 @@ export function distanceInWordsToNow(date) {
...
+  return parseInt(expires_at, 10) < Date.now()
+}
+
+/**
+ * Returns if object is Open
+ * @param {object} - obj with status & tx_status fields
+ * @param  {string} - status
+ * @returns {boolean}
+ */
+export function isOpen(obj, status) {
```
NicoSantangelo: I don't think this is necessary? I really like `hasStatus`, but the rest of the codebase ended up using `isOpen(publication, STATUS.open)` which means that I can do `isOpen(publication, STATUS.cancelled)` and it won't really be true.

So, to me, just having hasStatus is enough.

webapp/src/modules/mortgage/reducer.js
```diff
@@ -0,0 +1,66 @@
...
+    default:
+      return state
+  }
+}
+
+export default combineReducers({
+  data,
+  loading,
+  error
+})
```
NicoSantangelo: I like this approach! buuut (sorry 😭) I'm not sure about having it just here. Every other reducer uses only one function.
@cazala I'll leave the veredict to you.

webapp/src/modules/mortgage/sagas.js
```diff
@@ -0,0 +1,143 @@
...
+        loanMetadata // Metadata
+      )
+    )
+
+    // Sign the loan
+    let { signature } = yield call(() => eth.sign(loanIdentifier))
+    const approveSignature = signature.slice(2)
+    let r = `0x${approveSignature.slice(0, 64)}`
+    let s = `0x${approveSignature.slice(64, 128)}`
+    let v = eth.utils.toDecimal(`0x${approveSignature.slice(128, 130)}`)
```
NicoSantangelo: Houston we (may) have a problem here. Checking the implementation on `decentraland-eth` it looks like `LedgerWallet#sign` has a different result from `NodeWallet#sign`.

Have you tried it with a ledger? If you do have one, remember that it only works in https! so you may want to use https://ngrok.com/ to bridge the connection. @abarmat can help you test too.

I would say that, *if the results are incompatible*, we should change `decentraland-eth`s implmentation. But if we're in a hurry we can typecheck the wallet here and make the change in the meantime.

webapp/src/modules/mortgage/sagas.js
```diff
@@ -0,0 +1,143 @@
...
+    yield put(push(locations.activity))
+  } catch (error) {
+    yield put(cancelMortgageFailure(error.message))
+  }
+}
+
+function* handleFetchActiveParcelMortgagesRequest(action) {
+  try {
+    const { x, y } = action
+    const mortgages = yield call(() => api.fetchActiveMortgages(x, y, 'active'))
```
NicoSantangelo: Repeating myself but, same comment as the API, I would expect `api.fetchActiveMortgages(x, y)` or `api.fetchMortgages(x, y, 'active')`

webapp/src/modules/parcels/selectors.js
```diff
@@ -1,7 +1,11 @@
 import { createSelector } from 'reselect'
-import { EDIT_PARCEL_REQUEST } from './actions'
+import { EDIT_PARCEL_REQUEST, FETCH_PARCEL_REQUEST } from './actions'
 import { getPublications as getAllPublications } from 'modules/publication/selectors'
 import { buildCoordinate } from 'lib/utils'
+import { isLoadingType } from 'modules/loading/selectors'
+import { getMappedMortgages } from 'modules/mortgage/selectors'
+// import { getAddress } from 'modules/wallet/selectors'
```
NicoSantangelo: Whops commented out code

webapp/src/modules/mortgage/selectors.js
```diff
@@ -0,0 +1,32 @@
...
+import { buildCoordinate } from 'lib/utils'
+
+export const getState = state => state.mortgages
+export const getData = state => getState(state).data
+export const getLoading = state => getState(state).loading
+export const getMortgages = state => getData(state).mortgages
+export const isFetchingParcelMortgages = state =>
+  isLoadingType(getLoading(state), FETCH_ACTIVE_PARCEL_MORTGAGES_REQUEST)
+export const getMappedMortgages = state =>
+  Object.keys(getMortgages(state)).map(key => getMortgages(state)[key])
```
NicoSantangelo: I don't know if `Mapped` is the right word here (I had to check the implementation to get what it was doing). Maybe `getMortgagesArray`? (I know it's not _that_ different)

webapp/src/modules/wallet/sagas.js
```diff
@@ -73,22 +89,28 @@ function* handleConnectWalletRequest(action = {}) {
...
 
     const [
       network,
       balance,
       ethBalance,
       approvedBalance,
-      isLandAuthorized
+      isLandAuthorized,
+      isMortgageApprovedForMana,
+      isMortgageApprovedForRCN
```
NicoSantangelo: maybe these two should be called `mortgageManaAllowance` and `mortgageRCNAllowance` ? so then 
      isMortgageApprovedForMana: isMortgageApprovedForMana > 0,
      isMortgageApprovedForRCN: isMortgageApprovedForRCN > 0

nachomazzara: Yes, it is better

webapp/src/modules/parcels/selectors.js
nachomazzara: Oooohhh!!! Thanks you noticed. It was a selector imported as undefined. I forgot to revisited

webapp/src/modules/mortgage/reducer.js
nachomazzara: I told @cazala about it. He was ok

webapp/src/components/BuyParcelByMortgagePage/MortgageForm/MortgageForm.js
nachomazzara: I did it because I think it is better than having multiple handlers doing the same for each input

src/Translation/locales/es.json
nachomazzara: Ohhh 🤕 translation script. Will take a look

src/Mortgage/Mortgage.router.js
nachomazzara: Well, I hace the same feeling when doing it.

Yes, The endpoint handler should be named as `getMortgagesInCoordinate`

About `status`, although`active` is not a valid status for mortgages, will let us use the request without `status` to retrieve all mortgages.

NicoSantangelo: > About status, althoughactive is not a valid status for mortgages, will let us use the request without status to retrieve all mortgages. E.g: if we want to get a mortgages history. Dashboard data, etc

True! but we can also accomplish that by keeping the `active` abstraction in the front-end and supporting an array of status on this request. Which will work for future references and it's easier to make public (we won't have to explain what active means)

src/Translation/locales/es.json
NicoSantangelo: Ooh something told me that it came from there :(

webapp/src/modules/mortgage/reducer.js
NicoSantangelo: Yeap me too!, my problem was opening this file and encountering a different implementation from all the other reducers.
I don't want to be rigid tho, but sometimes I prefer keeping an slighty more cumbersome implementation if that means easier induction of new people to the project and removing the possibility of having yet another implementation later.

That said, 👍

src/Mortgage/Mortgage.router.js
nachomazzara: Im not against top level abstractions which help users memo ways to use api. Said that, Im ok to support status array, but also the other way.

Anyways, so far lets go with array support only

webapp/src/modules/mortgage/reducer.js
nachomazzara: Thanks, I would like To propose this “new” way to architecting reducers if we all agree obviusly

webapp/src/components/MortgageActions/MortgageActions.container.js
cazala: ☝️

webapp/src/components/Parcel/Parcel.js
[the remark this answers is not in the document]
cazala: @NicoSantangelo @nachomazzara maybe not for this PR, but I would rather have an ```onAccess: (parcel, wallet) => boolean``` callback that receives parcel and wallet and return a boolean (either the user can access or not) and then let each component that uses `<Parcel>` be the owner of that business logic, otherwise we keep adding rules to `<Parcel>` that are too use-case-specific.

webapp/src/lib/utils.js
cazala: +1 ☝️

webapp/src/modules/mortgage/sagas.js
cazala: ☝️

webapp/src/modules/mortgage/selectors.js
cazala: ☝️ @nachomazzara @NicoSantangelo **watch out**, this is missing `reselect`! 

A new instance of the array is created every time the selector is called. 

This is used by another selector `getMortgagedParcels` in `modules/parcels/selectors`, and this will kill that selectors memoization too.

`getMortgagedParcels` is used by `getAddresses` which is used by `getWallet`(both wont be able to memoize their results).

`getWallet` is used by 13 containers, and their `mapStateToProps` gets called on every `dispatch`. Since the selectors they use are not memoized, the `shallowCompare` they do will return false, triggering renders down the React tree (not only the containers, also their children if they receive the non-memoized prop).

webapp/src/modules/mortgage/utils.js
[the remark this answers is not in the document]
cazala: can we use `hasStatus` ?

webapp/src/modules/parcels/selectors.js
cazala: ☝️

webapp/src/modules/wallet/sagas.js
cazala: ☝️

PR #224, opened 2018-05-17
feat: create estate UI

fmiras:
- [x] Location:  +2 locations
- [x] Routes: +1 route
- [x] Component: Page -> (use ParcelDetailPage as layout model)
- [x] Map: onClick and selected
- [x] Validate that is more than 1 parcel
- [x] Validate that selected parcels are from the operating owner
- [x] Create estate modal and form to edit name and description
- [x] Calculate center and zoom of estate preview (don't use first parcel X and Y)
- [x] Validate that selected parcels are neighbours (pending decision)
- [x] Create actions, reducers and saga
- [x] Show estates on MY LAND
- [x] Create Estate PropType
- [x] Connect with Estate Backend
- [x] Optionally enable Estates feature
- [x] Translations


fmiras:
Preview of estate: 
screen shot 2018-05-21 at 4 20 49 pm

src/Asset/utils.js
```diff
@@ -0,0 +1,13 @@
+import { SQL } from '../database'
+import { PublicationQueries } from '../Publication'
+
+export const getFindByOwnerQuery = async (owner, table) =>
```
NicoSantangelo: Even though I left another approach on `findByOwner`, the async is not necessary here

src/Asset/Asset.js
```diff
@@ -10,11 +11,7 @@ export class Asset {
   }
 
   async findByOwner(owner) {
-    return db.query(SQL`SELECT ${SQL.raw(this.tableName)}.*, (
-        ${PublicationQueries.findLastParcelPublicationJsonSql()}
-      ) as publication
-        FROM ${SQL.raw(this.tableName)}
-        WHERE owner = ${owner}`)
+    return db.query(await getFindByOwnerQuery(owner, this.tableName))
```
NicoSantangelo: I think we can get away without a `utils` here. The Estates query will need publications as well so maybe we can do something like this:

```javascript
async findByOwner(owner) {
    return await db.query(SQL`SELECT ${SQL.raw(this.tableName)}.*, (
        ${AssetQueries.findLastPublicationJsonSql(this.tableName)}
      ) as publication
        FROM ${SQL.raw(this.tableName)}
        WHERE owner = ${owner}`)

File: src/Asset/utils.js (hunk 1 above)
Comment (fmiras) [reply src/Asset/utils.js]: Yeah, that was my bad 😅 I firstly made this function return `db.query` instead of just the string SQL statement, forgot to remove `async/await`. I'll remove it

File: src/Asset/Asset.js (hunk 2 above)
Comment (fmiras) [reply src/Asset/Asset.js]: I can rename `utils.js` to `AssetQueries` but my idea was to take the same approach as this one. According to the publications, yes, we're probably going to need them to estates to but there isn't a first `Auction` transaction as same as the estates right? Correct me if I'm wrong but doesn't the publications stands for a "parcel instance"?

Comment (fmiras) [reply src/Asset/Asset.js]: Already understood! what do you think on https://github.com/decentraland/marketplace/pull/224/commits/e43187928058505515865d9e692de7e890cf7ff2?

File: webapp/src/components/CreateEstatePage/CreateEstatePage.js [hunk 3]
```diff
@@ -0,0 +1,77 @@
...
+
+  handleChangeParcels = parcels => {
+    const { estate } = this.state
+    this.setState({
+      estate: {
+        ...estate,
+        data: {
+          ...estate.data,
+          parcels
+        }
```
nachomazzara: I can not see the usage of `...estate`  before `data` based on `initialState`
Maybe this could be enough
```
this.setState({
   estate: {
     data: {
         ...estate.data,
         parcels
      }
   }
})
```

webapp/src/components/CreateEstatePage/CreateEstatePage.js
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  handleSubmit = () => {
+    this.props.createEstate(this.state.estate)
+  }
+
+  render() {
+    const { isSelecting, estate } = this.state
+
+    return isSelecting ? (
```
nachomazzara: I would simplify this with 

```
return (
      <React.Fragment>
 isSelecting ?
        <EstateSelect
          value={estate.data.parcels}
          onContinue={this.handleSwitch}
          onChange={this.handleChangeParcels}
        />
     :  <EditEstate
          value={estate}
          onCancel={this.handleSwitch}
          onChange={this.handleChange}
          onSubmit={this.handleSubmit}
        />
      </React.Fragment>
    )
```

webapp/src/components/CreateEstatePage/EstateSelect/EstateSelect.container.js
```diff
@@ -0,0 +1,28 @@
...
+import { navigateTo } from 'modules/location/actions'
+import { getError, isLoading } from 'modules/parcels/selectors'
+import EstateSelect from './EstateSelect'
+
+const mapState = (state, ownProps) => {
+  const { x, y } = getMatchParams(ownProps)
+  return {
+    x: parseInt(x, 10),
+    y: parseInt(y, 10),
+    isLoading: isLoading(state),
```
nachomazzara: Oh, I think `isLoading` is not used

webapp/src/components/CreateEstatePage/EstateSelect/EstateSelect.js
```diff
@@ -0,0 +1,134 @@
...
+    x: PropTypes.number.isRequired,
+    y: PropTypes.number.isRequired,
+    error: PropTypes.string,
+    onCancel: PropTypes.func.isRequired,
+    onContinue: PropTypes.func.isRequired,
+    onChange: PropTypes.func.isRequired,
+    value: PropTypes.arrayOf(coordsType).isRequired
+  }
+
+  isNeighbour = (x, y) => coords => {
```
nachomazzara: What do you think of moving these methods to something an utils file like `stateUtils.js`?

webapp/src/components/CreateEstatePage/EstateSelect/EstateSelect.js
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  handleParcelClick = wallet => (x, y) => {
+    if (!isOwner(wallet, x, y)) {
+      return
+    }
+
+    if (!this.hasNeighbour(x, y)) {
+      return
+    }
```
nachomazzara: Maybe we can merge these two `if` into one

```
if (!isOwner(wallet, x, y) || !this.hasNeighbour(x, y)) {
       return false
}
```

Also, I would prefer to return `false` instead of `undefined`

webapp/src/components/CreateEstatePage/EstateSelect/EstateSelect.js
```diff
@@ -0,0 +1,134 @@
...
+      const newParcels = parcels.filter(
+        coords => !(coords.x === x && coords.y === y)
+      )
+
+      if (!this.areConnected(newParcels) && newParcels.length > 1) {
+        return
+      }
+
+      onChange(newParcels)
+      return
```
nachomazzara: why not `return Onchange(...)`

webapp/src/components/ProfilePage/ProfilePage.container.js
```diff
@@ -22,11 +22,13 @@ const mapState = (state, { location, match }) => {
   let parcels = []
   let contributions = []
   let publishedParcels = []
+  let estates = []
   let mortgagedParcels = []
   if (address in addresses) {
     parcels = addresses[address].parcels
     contributions = addresses[address].contributions
     publishedParcels = addresses[address].publishedParcels
+    estates = addresses[address].estates.filter(e => e.data.parcels.length > 0)
```
nachomazzara: What do you think of moving this logic inside the `getAddreses` selector

webapp/src/components/CreateEstatePage/EditEstate/EditEstateForm/EditEstateForm.js
```diff
@@ -0,0 +1,91 @@
...
+
+    return (
+      <Form className="EditEstateForm" onSubmit={preventDefault(onSubmit)}>
+        <Form.Field>
+          <label>{t('estate_edit.name')}</label>
+          <Input
+            type="text"
+            value={name}
+            onChange={this.handleNameChange}
+            error={name.length > 50}
```
NicoSantangelo: Now that this two hardcoded values (`50` and `140`) are repeated here and on `EditParcelForm` can we abstract them into a method/constant?

Maybe `lib/utils` or `lib/dataUtils` and `isValidName`+`isValidDescription`

webapp/src/components/CreateEstatePage/EditEstate/EditEstateForm/EditEstateForm.js
```diff
@@ -0,0 +1,91 @@
...
+          />
+        </Form.Field>
+        <Form.Field>
+          <label>{t('estate_edit.description')}</label>
+          <Input
+            type="text"
+            value={description}
+            onChange={this.handleDescriptionChange}
+            error={name.length > 140}
```
NicoSantangelo: This should be `description.lenght`. We have the same bug on `EditParcelForm`, please change it there too ! :P

webapp/src/components/CreateEstatePage/EditEstate/EstateModal/EstateModal.js
```diff
@@ -0,0 +1,97 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { Button, Grid, Header } from 'semantic-ui-react'
+import ParcelPreview from 'components/ParcelPreview'
+import { coordsType } from 'components/types'
+import { t } from 'modules/translation/utils'
+import './EstateModal.css'
+import { calculateZoomAndCenter } from 'modules/estates/utils'
```
NicoSantangelo: Nitpick mode on: Can we reorder this so `import './EstateModal.css'` is the last import? for consistency sake

webapp/src/components/CreateEstatePage/EditEstate/EstateModal/EstateModal.js
```diff
@@ -0,0 +1,97 @@
...
+                  {confirmLabel || t('global.confirm')}
+                </Button>
+              </Grid.Column>
+            </div>
+          )}
+        </div>
+      </div>
+    )
+  }
+}
```
NicoSantangelo: Maaaaybe we can abstract some of this logic to a `AssetModal`? because I think there's a lot of repeated logic between this and `ParcelModal`

NicoSantangelo: Maaaaybe we can abstract some of this logic to a `AssetModal`? because I think there's a lot of repeated logic between this and `ParcelModal`

What do you think?

webapp/src/components/CreateEstatePage/EstateSelect/EstateSelect.js
```diff
@@ -0,0 +1,134 @@
...
+        return (
+          parcels.some(
+            coords2 => coords.x === coords2.x && coords.y === coords2.y
+          ) &&
+          !alreadyTraveled.some(
+            coords2 => coords.x === coords2.x && coords.y === coords2.y
+          )
+        )
+      }
+    )
```
NicoSantangelo: This is a bit confusing to me, calling reading the code I don't see what `neighbours` is at first glance. Can we maybe name this entire method?

```javascript
this.getNeighbours(actual.x, actual.y, parcels).filter(
      coords => {
        return (
          parcels.some(
            coords2 => coords.x === coords2.x && coords.y === coords2.y
          ) &&
          !alreadyTraveled.some(
            coords2 => coords.x === coords2.x && coords.y === coords2.y
          )
        )
      }
    )
```

webapp/src/components/CreateEstatePage/EstateSelect/EstateSelect.js
```diff
@@ -0,0 +1,134 @@
...
+    if (!this.hasNeighbour(x, y)) {
+      return
+    }
+
+    const { value: parcels, onChange } = this.props
+    const isSelected = parcels.some(coords => coords.x === x && coords.y === y)
+
+    if (isSelected) {
+      const newParcels = parcels.filter(
+        coords => !(coords.x === x && coords.y === y)
```
NicoSantangelo: I think this Component might benefit from having a method that compares parcels somewhere, to stop repeating this logic: `coords.x === x && coords.y === y` and to be able to negate: `!isEqual(parcel1, parcel2)` clearly

There are a few more places where a check like is performed like Parcel.container.js and ParcelCanvas.js

webapp/src/components/ParcelTags/ParcelTags.js
```diff
@@ -21,6 +22,10 @@ export default class ParcelTags extends React.PureComponent {
     showDetails: false
   }
 
+  isParcel() {
+    return this.props.parcel
```
NicoSantangelo: I'd return `!!this.props.parcel` here, to comply to the `isXXX returns boolean` convention

webapp/src/components/ParcelTags/ParcelTags.js
```diff
@@ -65,8 +70,35 @@ export default class ParcelTags extends React.PureComponent {
...
+        if (finalProximity[key].distance > actualProximity[key].distance) {
+          finalProximity[key] = actualProximity[key]
+          return
+        }
+      })
+      return finalProximity
+    }, {})
+
+    return { proximity }
+  }
```
NicoSantangelo: _Not imperative_ but maybe this method could look like this?

webapp/src/components/ParcelTags/ParcelTags.js
```diff
@@ -65,8 +70,35 @@ export default class ParcelTags extends React.PureComponent {
...
+
```
NicoSantangelo: _Not imperative_ but maybe this method could look like this?

```
getTags() {

Comment (NicoSantangelo) [inline webapp/src/components/ParcelTags/ParcelTags.js]: _Not imperative_ but maybe this method could look like this?

```
getTags() {
  if (isParcel()) {
     return this.getParcelTags()
  } else {
     return this.getStateTags()
  }
```

just to prepare the reader for each method definition.
Again I'm not 100% sure it's the way to go

File: webapp/src/modules/address/reducer.js [hunk 19]
```diff
@@ -136,6 +156,26 @@ export function addressReducer(state = INITIAL_STATE, action) {
...
+      )
+
+      return {
+        ...state,
+        data: {
+          ...state.data,
+          [owner]: {
+            ...state.data[owner],
+            parcel_ids: [
+              ...state.data[owner].parcel_ids.filter(x => !parcelIds.has(x))
```
NicoSantangelo: filter already returns a new array so the `[...]` is not necessary

webapp/src/modules/estates/sagas.js
```diff
@@ -0,0 +1,73 @@
...
+import { api } from 'lib/api'
+import { buildCoordinate } from 'lib/utils'
+
+function validateCoords(x, y) {
+  if (!inBounds(x, y)) {
+    throw new Error(`Coords (${x}, ${y}) are outside of the valid bounds`)
+  }
+}
+
+// delete when estate contract returns an address
```
NicoSantangelo: Can we add a `TODO: ` here?

webapp/src/modules/estates/sagas.js
```diff
@@ -0,0 +1,73 @@
...
+  } catch (error) {
+    console.warn(error)
+    yield put(fetchEstateFailure(id, error.message))
+  }
+}
+
+export function* estateSaga() {
+  yield takeEvery(CREATE_ESTATE_REQUEST, handleCreateEstateRequest)
+  yield takeEvery(FETCH_ESTATE_REQUEST, handleEstateRequest)
+}
```
NicoSantangelo: _Nitpick again_: To keep the files consistent, I'd move this function to the top of the file, below the imports

webapp/src/components/CreateEstatePage/EditEstate/EditEstate.css
```diff
@@ -0,0 +1,4 @@
+.EditEstate .subtitle {
+  color: white;
```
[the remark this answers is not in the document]
NicoSantangelo: `#FFF` :P

src/Asset/Asset.js
```diff
@@ -10,15 +10,15 @@ export class Asset {
   }
 
   async findByOwner(owner) {
-    return await db.query(SQL`SELECT ${SQL.raw(this.tableName)}.*, (
-        ${PublicationQueries.findLastParcelPublicationJsonSql()}
+    return db.query(SQL`SELECT ${SQL.raw(this.tableName)}.*, (
```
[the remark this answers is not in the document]
fmiras: fixed 👍 3e45c41b2a48683d58c217acf05acc6e1efa2f07

src/Mortgage/Mortgage.router.js
```diff
@@ -53,14 +53,14 @@ export class MortgageRouter {
   async getMortgagesByBorrower(req) {
     const borrower = server.extractFromReq(req, 'address')
     const status = server.extractFromReq(req, 'status')
-    return await Mortgage.findByBorrower(borrower, status)
+    return Mortgage.findByBorrower(borrower, status)
```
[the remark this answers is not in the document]
fmiras: As we talked earlier, the expression `return await` is redundant since we are inside an async function unless we are inside a try/catch block and we need to resolve the promise before to know if it has any throwers

webapp/src/components/CreateEstatePage/CreateEstatePage.container.js
```diff
@@ -0,0 +1,30 @@
...
+import { createEstateRequest } from 'modules/estates/actions'
+import { getError, isLoading } from 'modules/parcels/selectors'
+import CreateEstatePage from './CreateEstatePage'
+
+const mapState = (state, ownProps) => {
+  const { x, y } = getMatchParams(ownProps)
+  return {
+    x,
+    y,
+    isLoading: isLoading(state),
```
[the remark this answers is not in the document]
fmiras: True! I may forget to remove them after I stopped using them. fixed ac5f3184460e6c5dc720bf2b10b9ec99c87de2b7!

webapp/src/components/CreateEstatePage/EditEstate/EditEstate.css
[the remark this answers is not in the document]
fmiras: I also prefer to use hexa as well, the problem is that I took as a following model the component `EditParcelPage` that is using `color: white;` in it's own .css file. What should we prioritize, consistency or color accuracy?

webapp/src/components/CreateEstatePage/CreateEstatePage.js
fmiras: Right know it's useless, before it was being use because the estate object was having some more attributes in addition to data object. fixed 14e24c4

webapp/src/components/CreateEstatePage/CreateEstatePage.js
fmiras: nice idea! fixed 94aefd8

webapp/src/components/CreateEstatePage/EstateSelect/EstateSelect.container.js
fmiras: True! removed 2e2adc9

webapp/src/components/CreateEstatePage/EditEstate/EditEstate.css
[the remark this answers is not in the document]
nachomazzara: I prefer consistency and when we see color by names replace them by hexa, @NicoSantangelo @cazala thoughts?

webapp/src/components/CreateEstatePage/EstateSelect/EstateSelect.js
fmiras: Great idea! Will do after the changes review of this same file :)

webapp/src/components/CreateEstatePage/EstateSelect/EstateSelect.js
fmiras: Actually, here's my bad, we shouldn't return anything at all, since I'm using return statements to break the execution on non-passed validations, will remove the last return statement.

webapp/src/components/CreateEstatePage/EstateSelect/EstateSelect.js
fmiras: I don't know if that's the best way for making function validations, the thing is that if we change the 2 ifs statements to one it will execute both of the validations in vain, because if the first one already make the function to not execute anything is a little bit weird to make another validation. What's your opinion on this?

webapp/src/components/ProfilePage/ProfilePage.container.js
fmiras: Nice! done 0174c65

webapp/src/components/CreateEstatePage/EstateSelect/EstateSelect.js
fmiras: done 0174c65

webapp/src/components/ProfilePage/ProfilePage.container.js
fmiras: Nice! done e7d47b1

webapp/src/components/CreateEstatePage/EditEstate/EditEstateForm/EditEstateForm.js
fmiras: Good idea! done 72f4a70

webapp/src/components/CreateEstatePage/EditEstate/EditEstateForm/EditEstateForm.js
fmiras: Same here, fixed on 72f4a70

webapp/src/components/CreateEstatePage/EditEstate/EstateModal/EstateModal.js
fmiras: Agree on this, moved 3b80f5b

webapp/src/components/CreateEstatePage/EstateSelect/EstateSelect.css
```diff
@@ -0,0 +1,65 @@
...
+
+.ParcelDetail .parcel-actions-container {
+  display: flex !important;
+  flex-flow: row nowrap;
+  justify-content: flex-end;
+  align-items: center;
+}
+
+.ParcelDetail h3 {
+  font-family: var(--font-family) !important;
```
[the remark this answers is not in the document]
fmiras: No, I've already remove the extra css here 585db36

webapp/src/components/CreateEstatePage/EditEstate/EstateModal/EstateModal.js
fmiras: We can, yes I would wait for the Estate feature to finish because we have to add some more things here such as parcels preview, parcel remove, edit buy and sell parcel so I started just copying the logic of `ParcelModal` to later remove unused stuff and finally merge logic into an `AssetModal`, does that sound good for you?

webapp/src/components/CreateEstatePage/EstateSelect/EstateSelect.js
fmiras: I call `neighbour` to a parcel that the user also own and at the same time is next to another parcel, do you have any alternative to propose?

webapp/src/components/ParcelTags/ParcelTags.js
fmiras: I'm not sure either, but done 9d2c76c :)

webapp/src/modules/address/reducer.js
fmiras: True! removed 539d617

webapp/src/modules/estates/sagas.js
fmiras: done 539d617!

webapp/src/components/CreateEstatePage/EstateSelect/EstateSelect.js
fmiras: I agree to merge the function validations with and OR, in some point maybe a very big or concatenated chain may be difficult to read, but in this case there are only two.

According to the boolean return, why do you prefer returning false instead of nothing? The purpose of this return statement is to stop executing code because is conditional code, also I think that returning a boolean flag will make harder to understand because coders may think that is meant to be use as a function for validating instead of execution.

nachomazzara: IMHO, it is mostly for consistency and code design. If you need to call this function later and wait for a response, you already have a value (`false`) instead of `undefined`.

fmiras: The thing is that the function must not be called at all for waiting a response, and changing the function signature from void to boolean is going to be a patch for bad usage cases. Although we are not explicitly defining the signature of the function, it exists and we have to respect it. Unless we need to logical change the signature to return a boolean value for a concrete situation, I stand to keep returning anything, because isn't a return, is used as `break`

webapp/src/components/CreateEstatePage/EstateSelect/EstateSelect.js
fmiras: Added a TODO comment eebd404

webapp/src/modules/estates/utils.js
```diff
@@ -0,0 +1,17 @@
+export const calculateZoomAndCenter = parcels => {
+  const xs = [...new Set(parcels.map(coords => coords.x).sort())]
+  const ys = [...new Set(parcels.map(coords => coords.y).sort())]
+  const x = xs[parseInt(xs.length / 2, 10)]
+  const y = ys[parseInt(ys.length / 2, 10)]
+  const center = { x, y }
+  const zoom = 1 / (xs.length + ys.length) * 7.5
```
cazala: Magic number :P ? I haven't tested this branch yet, I'll check it out later, but does this behave well for any estate size? I think we should be able compute the `size` prop instead of the `zoom`, we can do that easily if we have the `width`, `height` and `parcels`, which we should have from any `EstatePreview`.  Most of the code around the map is my fault :P so I'll help you out, no worries.

webapp/src/components/CreateEstatePage/EstateSelect/EstateSelect.js
nachomazzara: didn't know about that guidelines

webapp/src/components/CreateEstatePage/EstateSelect/EstateSelect.js
eordano: contiguous

src/Translation/locales/es.json
```diff
@@ -125,14 +140,12 @@
     "duration_placeholder": "Duración en días",
     "errors": {
       "amount_minimum": "La cantidad debe ser mayor que {amount}",
-      "duration_gte_payable":
-        "La duración debe ser mayor o igual que la paga en días",
```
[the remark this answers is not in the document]
fmiras: done f29e7b2!

src/Translation/locales/en.json
```diff
@@ -47,6 +47,20 @@
     "invalid":
       "You have set a wrong derivation path for your wallet. Your derivation path must follow pattern 44'/60|61'/x'/n"
   },
+  "estate": {
+    "acquired_at": "Acquired at {date}",
```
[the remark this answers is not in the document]
fmiras: I think that's a good idea! done f29e7b2

PR #235, opened 2018-05-23
feat: show ongoing mortgages

nachomazzara:
## Cards
- [Show claimed/started mortgages](https://trello.com/c/aMCznz5D/299-show-claimed-mortgages)
- [Show pending mortgages not expired and with open publications](https://trello.com/c/g6DxrXpR/298-do-not-show-active-mortgages-if-publication-is-cancelled-or-finished)
- [Missing transaction history at parcel detail ](https://trello.com/c/GtqLxWcT/316-missing-transactions-history-at-parcel-detail-remove-assetid-from-blacklist-property-to-omit-at-publication)
- [Refactor utils for assets status (publication and mortgage)](https://trello.com/c/KkJHC7Fd/317-refactor-utils-for-assets-status-publication-and-mortgage)
- [Show mortgage tab at profile only if owner](https://trello.com/c/0Sg0WsVE/318-show-mortgage-tab-at-profile-only-if-owner)

## How to test it
- You must truncate mortgages table cause I added some new `not null` fields without default and run again the monitor:

`npm run cli truncate mortgages`
`PROCESS_EVENTS_DELAY=5000 npm run monitor-blockchain -- --from-block 2000000 --watch                  `

## Notes

- `"eslint-plugin-react": "7.4.0",` to avoid react 16.3 lint issues

# Important!!!

Better to look [this PR](https://github.com/decentraland/marketplace/pull/244) has both changes


src/Parcel/Parcel.spec.js
```diff
@@ -106,19 +107,55 @@ describe('Parcel', function() {
...
+        price: 1500,
+        expires_at: new Date().getTime() * 1000,
+        contract_id: '0xdseadbeef',
+        block_time_created_at: null,
+        block_time_updated_at: null,
+        marketplace_id: '0x113322',
+        block_number: 1
+      }
+      await Publication.insert(publication)
+      await Publication.insert(publication2)
```
fmiras: Can we run this two with `Promise.all(<PromisesArray>)` since we are not assigning them to variables we can take advantage of asynchronism right?

src/Parcel/Parcel.spec.js
```diff
@@ -106,19 +107,55 @@ describe('Parcel', function() {
   describe('.findWithLastActiveMortgageByBorrower', function() {
     beforeEach(() => new ParcelService().insertMatrix(0, 0, 10, 10))
 
-    it('should return parcels with mortgages open/claimed by borrower', async function() {
```
fmiras: I saw in other spec files we are using arrow functions, this file is using es5 ones, is that for any particular reason? If not I think is better to be consistent across all the other code.

src/Publication/Publication.spec.js
```diff
@@ -81,12 +82,8 @@ describe('PublicationService', function() {
   describe('#getModelByType', function() {
     it('should return the model class for the supplied type', function() {
```
fmiras: Same here with the function/arrow inconsistency maybe we can do the opposite and make all the functions use the es5 one instead of arrows

webapp/package.json
```diff
@@ -26,7 +26,7 @@
     "eslint": "4.10.0",
     "eslint-config-prettier": "^2.6.0",
     "eslint-plugin-prettier": "^2.3.1",
-    "eslint-plugin-react": "^7.5.1",
+    "eslint-plugin-react": "7.4.0",
```
fmiras: Great! Maybe not for this PR, but what do you guys think about pinning up dependencies version and using a tool like greenkeeper.io so we don't have new uncontrolled errors and we can control version upgrades? @abarmat @cazala @NicoSantangelo

webapp/src/modules/publication/sagas.js
```diff
@@ -35,6 +35,7 @@ export function* publicationSaga() {
 
 function* handlePublicationsRequest(action) {
   try {
+    console.log(action)
```
fmiras: I think you forgot to remove this!

src/Asset/Asset.spec.js
```diff
@@ -128,7 +129,7 @@ describe('Asset', function() {
             price: 40,
             buyer: null,
             status: Publication.STATUS.open,
-            type: Publication.TYPES.parcel,
+            type: ASSET_TYPE.parcel,
```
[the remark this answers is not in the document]
nachomazzara: We had a conversation with @NicoSantangelo about it. One of the key concepts of what we don't have `Parcel.type` or `Entity.type` is because we could have dynamic entities in the future creating by an admin and it would be hard to maintain something like `DynamicEntity.TYPE`

src/Mortgage/Mortgage.spec.js
```diff
@@ -4,32 +4,44 @@ import { txUtils } from 'decentraland-eth'
...
 import { Mortgage } from '../Mortgage'
 import { Parcel } from '../Parcel'
+import { ASSET_TYPE } from '../Asset'
+import { Publication } from '../Publication'
 
 describe('Mortgage', function() {
+  const expires_at = new Date().getTime() * 1000
+  const is_due_at = 15000
+  const payable_at = 1000
+  let mortgage = {}
```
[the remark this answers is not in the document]
nachomazzara: I did what you said on the [`beforeEach` method](https://github.com/decentraland/marketplace/pull/235/files/d3611929714aed9e0e2425c641bfcc4da81903b0#diff-b472fc2ffbb046805ee9a677076787ebR16)

src/Mortgage/Mortgage.spec.js
```diff
@@ -4,32 +4,44 @@ import { txUtils } from 'decentraland-eth'
...
-        borrower: '0xdeadbeef33',
-        lender: null,
-        amount: 1500,
-        block_time_created_at: null,
-        block_time_updated_at: null,
-        block_number: 1,
-        is_due_at,
-        expires_at
-      }
       const mortgage2 = Object.assign({}, mortgage, {
```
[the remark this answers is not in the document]
nachomazzara: Yep, both work. I think I will keep `Object.asssign` (we are transpiling) it unless we find something against Obect.assign

src/Parcel/Parcel.spec.js
nachomazzara: Oh! yep, we are using both, we may create an issue or Trello card to make the change in another PR SQL.spec has arrow functions too

src/Mortgage/Mortgage.spec.js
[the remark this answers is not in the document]
fmiras: Yes but what I meant is that we don't need to use a `beforeEach` that redefines the object mortgage each time if we use on each test something like

src/Mortgage/Mortgage.spec.js
```diff
@@ -4,32 +4,44 @@ import { txUtils } from 'decentraland-eth'
...
+import { ASSET_TYPE } from '../Asset'
+import { Publication } from '../Publication'
 
 describe('Mortgage', function() {
+  const expires_at = new Date().getTime() * 1000
+  const is_due_at = 15000
+  const payable_at = 1000
+  let mortgage = {}
+
+  beforeEach(async () => {
```
fmiras: `async` isn't required here

src/Mortgage/Mortgage.spec.js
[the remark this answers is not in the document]
nachomazzara: Oh, I can't still see the advantage of not doing it in a test. Is it ok for you to keep the way it is?

src/Mortgage/Mortgage.spec.js
nachomazzara: good catch!!!

nachomazzara:
Recommend to review [this PR](https://github.com/decentraland/marketplace/pull/244) directly instead

webapp/src/modules/mortgage/utils.js
```diff
@@ -31,13 +31,17 @@ export function getLoanMetadata() {
...
 
 export function isMortgageActive(mortgage, parcel) {
-  return (
-    mortgage &&
-    parcel &&
-    ((isMortgagePending(mortgage) &&
-      isOpen(parcel.publication, PUBLICATION_STATUS.open)) ||
-      hasStatus(mortgage, [MORTGAGE_STATUS.ongoing, MORTGAGE_STATUS.paid]))
-  )
+  if (mortgage && parcel) {
```
fmiras: What if we first make validations and then the business logic? So we don't mix the operations with validations, in this case we can do something like
```js
if (!mortgage || !parcel) {
	return
}

// do stuff

src/Mortgage/Mortgage.spec.js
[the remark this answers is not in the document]
fmiras: The advantages are:
- Variables can still be called `mortgage` instead of `mortgage2` (code readability)
- We just build mortgages on tests that actually need it, and not all of them with a `beforeEach`
- Tests become more encapsulated since they wouldn't use an external estate (the `let mortgage = {}` variable)

PR #244, opened 2018-06-05
feat: paid mortgages & claim

nachomazzara:


scripts/monitor/processEvents.js
```diff
@@ -51,7 +52,7 @@ async function processNoParcelRelatedEvents(event) {
         log.info(`[${name}] Cancelling Mortgage ${_id}`)
         await Mortgage.update(
           {
-            status: Mortgage.STATUS.cancelled,
```
NicoSantangelo: While both `canceled` and `cancelled` are correct, the rest of the codebase uses `cancelled` for publication status. Can we change this one? just to be consistent

scripts/monitor/processEvents.js
```diff
@@ -66,6 +67,105 @@ async function processNoParcelRelatedEvents(event) {
...
+      break
+    }
+    case BlockchainEvent.EVENTS.partialPayment: {
+      const { _index } = event.args
+      const block_time_updated_at = await new BlockTimestampService().getBlockTime(
+        block_number
+      )
+      log.info(`[${name}] Partial Payment for loan ${_index}`)
+
+      const rcnEngineContract = await eth.getContract('RCNEngine')
```
NicoSantangelo: `getContract` is not `async`

scripts/monitor/processEvents.js
```diff
@@ -66,6 +67,105 @@ async function processNoParcelRelatedEvents(event) {
...
+      try {
+        log.info(`[${name}] Starting Mortgage ${_id}`)
+        await Mortgage.update(
+          {
+            status: Mortgage.STATUS.ongoing,
+            block_time_updated_at
+          },
+          {
+            mortgage_id: _id
+          }
```
NicoSantangelo: _Opinion_: I think that `{ mortgage_id: _id }` looks a bit better, same with `loan_id: _index` and the others below. If you agree and don't mind changing it, I'm all for it.

webapp/src/components/BuyParcelByMortgagePage/MortgageForm/MortgageForm.js
```diff
@@ -288,33 +285,18 @@ export default class MortgageForm extends React.PureComponent {
...
-          <Message icon>
-            <Icon name="circle notched" loading />
-            <Message.Content>
-              <TxStatus.Text
-                txHash={publication.tx_hash}
-                txStatus={publication.tx_status}
-              />
-            </Message.Content>
-          </Message>
-        ) : null}
```
NicoSantangelo: Why are we deleting the pending Message here? I'm not necessarily opposed but I think it can be useful.

Also, upon further instruction, `isIdle` marks if the transaction started but hasn't been confirmed by the user. For example if you leave the metamask popup open. It's not that important, but might be useful for disallowing new txs until the last one is confirmed

webapp/src/components/MortgageActions/MortgageActions.container.js
```diff
@@ -9,9 +16,18 @@ const mapState = (state, ownProps) => {
   }
 }
 
-const mapDispatch = (dispatch, { mortgage }) => ({
-  onCancel: () =>
-    dispatch(cancelMortgageRequest(mortgage.mortgage_id, mortgage.asset_id))
-})
+const mapDispatch = (dispatch, { mortgage }) => {
+  const coordinates = splitCoordinate(mortgage.asset_id)
```
NicoSantangelo: To be a bit more clear, we could change this to  `const [x, y] = splitCoordinate(asset_id)`

Also, looking at this I found a bug on `cancelMortgageSuccess` in `mortgage/actions.js`:

https://github.com/decentraland/marketplace/pull/244/files#diff-1bc520535984d0e30212dd309cfb59b1R46

```javascript
  const { x, y } = splitCoordinate(assetId)
// Should be
  const [ x, y ] = splitCoordinate(assetId)
```

webapp/src/components/ParcelCard/ParcelCard.css
```diff
@@ -2,11 +2,18 @@
...
   text-transform: capitalize;
+}
+
+.ParcelCard .Mortgage.pending{
   color: var(--orange);
 }
 
+.ParcelCard .mortgage-status.ongoing{
+  color: var(--light-blue);
+}
```
NicoSantangelo: The annoying co-worker strikes again. Can we prettify this file? to look like `.selector {` instead of `.selector{`

webapp/src/components/ParcelDetailPage/ParcelDetailPage.container.js
```diff
@@ -8,28 +8,24 @@ import { getDistricts } from 'modules/districts/selectors'
...
-      isLoading: isLoading(state),
-      error: getError(state),
-      districts: getDistricts(state),
-      publications: getParcelPublications(state),
-      mortgages: getParcelMortgages(state)
-    }
-  }
+  return state => ({
+    x,
+    y,
```
NicoSantangelo: Warn!

NicoSantangelo: ⚠️ 

I think this is re-introducing a bug where the `x,y` coordinates won't change and in consequence the detail won't reload when clicking the map

webapp/src/components/PayMortgagePage/PayMortgageForm/PayMortgageForm.js
```diff
@@ -0,0 +1,87 @@
...
+        )
+      })
+    }
+  }
+
+  handleSubmit = () => {
+    const {
+      mortgage: { loan_id, asset_id },
+      onSubmit
+    } = this.props
```
NicoSantangelo: _You don't need to change this_: this is perfectly valid, but I lean towards using something like:

```javascript
const { mortgage, onSubmit } = this.props
const { loan_id, asset_id } = morgage
```

it accomplishes the same in fewer lines and it's consistent with the rest of the codebase

webapp/src/components/PayMortgagePage/PayMortgagePage.container.js
```diff
@@ -0,0 +1,44 @@
...
+} from 'modules/mortgage/selectors'
+import { isLoading } from 'modules/parcels/selectors'
+import { getWallet, isConnected, isConnecting } from 'modules/wallet/selectors'
+
+const mapState = (state, ownProps) => {
+  const { x, y } = getMatchParamsCoordinates(ownProps)
+  const getMortgage = getParcelMortgageFactory(x, y)
+  return state => ({
+    x,
+    y,
```
NicoSantangelo: Same as `ParcelDetailPage` above, the `x,y` won't change if the reducer updates. Is this a problem here?

webapp/src/components/ParcelCard/ParcelCard.css
NicoSantangelo: Same goes for the other `.css` files, but then again, it's not really important

webapp/src/lib/utils.js
```diff
@@ -99,26 +99,25 @@ export function isExpired(expires_at) {
...
+ * @param  {array | string} - status
  * @returns {boolean}
  */
 export function hasStatus(obj, status) {
   return (
     obj &&
-    obj.status === status &&
-    obj.tx_status === txUtils.TRANSACTION_STATUS.confirmed &&
-    !isExpired(obj)
+    status.includes(obj.status) &&
```
NicoSantangelo: This shouldn't matter in the long run, buuut, if you have an status named `open` and another named `opened` then:

```javascript
  hasStatus({ status: 'open' }, 'opened') // true
```

I don't think it's a really a big deal, but wanted to point it out just in case

webapp/src/modules/mortgage/utils.js
```diff
@@ -21,11 +30,39 @@ export function getLoanMetadata() {
...
-      mortgage &&
-      mortgage.borrower === borrower &&
-      mortgage.status !== MORTGAGE_STATUS.cancelled
+export function isMortgageActive(mortgage, parcel) {
+  return (
+    mortgage &&
+    parcel &&
+    ((isMortgagePending(mortgage) &&
+      isOpen(parcel.publication, PUBLICATION_STATUS.open)) ||
+      hasStatus(mortgage, [MORTGAGE_STATUS.ongoing, MORTGAGE_STATUS.paid]))
```
NicoSantangelo: I think this is kind of confusing I can't see what it's doing at first glance. Can we maybe reword it?

Like

```javascript
  if (mortgage && parcel) {
    const isPending = isMortgagePending(mortgage) && isOpen(parcel.publication, PUBLICATION_STATUS.open)
    
    return isPending || hasStatus(mortgage, [MORTGAGE_STATUS.ongoing, MORTGAGE_STATUS.paid])
  }
```

(wrote on Github, don't take seriously)

webapp/src/modules/mortgage/utils.js
```diff
@@ -21,11 +30,39 @@ export function getLoanMetadata() {
...
+}
+
+/**
+ * Returns mortgages active -> pending status require an open publication
+ * @param {array} - mortgages
+ * @param  {array} - parcels
+ * @returns {array} - mortgages
+ */
+export function getActiveMortgages(mortgages, parcels) {
+  return mortgages
```
NicoSantangelo: if mortgages could be undefined, maybe instead of adding a `?:` we ca use:

`export function getActiveMortgages(mortgages = [], parcels = {}) {`

or 

export function getActiveMortgages(mortgages, parcels) {
  mortgages = mortgages || []

NicoSantangelo: if mortgages could be undefined, maybe instead of adding a `?:` we ca use:

`export function getActiveMortgages(mortgages = [], parcels = {}) {`

or 

```javascript
export function getActiveMortgages(mortgages, parcels) {
  mortgages = mortgages || []
```

scripts/monitor/processEvents.js
nachomazzara: 🤐

webapp/src/components/PayMortgagePage/PayMortgagePage.container.js
nachomazzara: It should not cause any problem here, no need for re-fetching

webapp/src/lib/utils.js
nachomazzara: Cool, what do you think of refactoring it once we need it?

PR #252, opened 2018-06-13
[WIP] feat: Show estates on map

fmiras:
## Tasks:
- [x] Enables estates on the map showing parcels with parcel color but connected as a district
- [x] Refactor of shared library where we split the logic of parcel utils onto a parcel one and a asset file so we reuse functions for estates
- [ ] Testing

## How to test
_pending test instructions_


shared/estate.js
```diff
@@ -0,0 +1,23 @@
...
+  return !!asset.parcels
+}
+
+export function getEstateByParcel(parcel, estates) {
+  return Object.keys(estates)
+    .map(estateId => estates[estateId])
+    .find(estate => estate.id === parcel.owner)
+}
+
+export function areOnSameEstate(parcels) {
```
NicoSantangelo: Maybe `belongToState` is a bit clearer?

shared/parcel.js
```diff
@@ -42,101 +8,8 @@ export function splitCoordinate(id) {
...
-    case TYPE.onSale:
-      return COLORS.onSale
-    case TYPE.unowned:
-      return COLORS.unowned
-    case TYPE.background:
-    default:
-      return COLORS.background
-  }
+export function isParcel(asset) {
+  return !!asset.x
```
NicoSantangelo: What if the `x` coordinate is `0`?

```

File: shared/parcel.js [hunk 3]
```diff
@@ -171,28 +44,30 @@ export function normalizeParcel(parcel, prevParcel = {}) {
...
   })
   return parcels
 }
 
+export function areConnected(parcels, parcelId, sideId) {
+  const parcel = parcels[parcelId]
+  const sideParcel = parcels[sideId]
+  return (
+    !sideParcel ||
+    isSameValue(parcel, sideParcel, 'district_id') ||
```
NicoSantangelo: I'm not convinced with this abstraction. It's not _really_ testing if two objects have the same value, it's testing that a property is not null in both _and_ it's the same.

I think it's only used here so I'm tempted to replace it with

```javascript
export function areConnected(parcels, parcelId, sideId) {
  const parcel = parcels[parcelId]
  const sideParcel = parcels[sideId]

  const sameDistrict = parcel.district_id === sideParcel.district_id
  const sameOwner = parcel.owner === sideParcel.ownwer

  return (
    !sideParcel ||
    (parcel.district_id && sameDistrict)
    (parcel.in_estate && parcel.owner && sameOwner) ||
  )
}
```

what do you think?

src/Map/Map.router.js
```diff
@@ -34,6 +31,7 @@ export class MapRouter {
       this.handleRequest(this.getParcelPNG)
     )
     // TODO: add an endpoint for Estates someday 🏌🏼‍
```
NicoSantangelo: I think we can delete the TODO! ⛳️

src/Map/Map.router.js
```diff
@@ -65,6 +63,27 @@ export class MapRouter {
...
+      const nw = server.extractFromReq(req, 'nw')
+      const se = server.extractFromReq(req, 'se')
+
+      const parcelsRange = await Parcel.inRange(nw, se)
+      const parcels = utils.mapOmit(parcelsRange, blacklist.parcel)
+      const estates = await EstateService.getByParcels(parcels)
+
+      const assets = { parcels, estates }
+      const total = parcels.length + estates.length
+      console.log(assets.estates)
```
NicoSantangelo: Whops

webapp/src/components/ParcelPreview/ParcelCanvas/ParcelCanvas.js
```diff
@@ -417,52 +402,29 @@ export default class ParcelPreview extends React.PureComponent {
...
-        publication,
-        connectedLeft: !!(parcel && parcel.connectedLeft),
-        connectedTop: !!(parcel && parcel.connectedTop),
-        connectedTopLeft: !!(parcel && parcel.connectedTopLeft),
-        color,
-        backgroundColor,
-        label,
-        description
-      }
+    const { wallet, parcels, districts, publications, estates } = this.props
```
NicoSantangelo: Why are we not caching anymore here? (the `this.cache[parcelId] = asset` is still there)

shared/estate.js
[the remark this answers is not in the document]
fmiras: Maybe, I don't like very much the name either but I wanted the function to explicity say the is validating that the parcels are all on the same estate, `belongToEstate` might be a little bit confusing, because I imagine that it validates that the parcels are on estates, but not at the same one, maybe `belongToSameEstate`?

src/Map/Map.router.js
fmiras: It's being deleted in another PR, because it refers to a `/estate` endpoint https://github.com/decentraland/marketplace/pull/265/files#diff-32768cada300df64cc722464f0ad95fbR30

shared/parcel.js
fmiras: Yes, my mistake was to force keep the `isSameValue`, I didn't even think if it was required or not, I'll put your implementation! (with typos fixed of course :P)

src/Map/Map.router.js
fmiras: nothing happened here 😅

shared/estate.js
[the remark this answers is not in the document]
NicoSantangelo: I like `belongToSameState`!

webapp/src/components/ParcelPreview/ParcelCanvas/ParcelCanvas.js
```diff
@@ -402,6 +402,10 @@ export default class ParcelPreview extends React.PureComponent {
   getParcelAttributes = (x, y) => {
     const parcelId = buildCoordinate(x, y)
 
+    if (this.cache[parcelId]) {
+      this.cache[parcelId]
```
NicoSantangelo: I think this is missing a return?

PR #268, opened 2018-06-18
chore: HTTP API docs

cazala:
In this PR I added a `docs/http-api.md` file with some documentation for our HTTP API (Mortgages docs are still missing) 

I made a few fixes and changes along the way:

- Removed the duplicated static constants `Publication.STATUS`, `Publication.TYPES`, `Mortgage.STATUS` and `Mortgage.TYPES`

- Added default values to the `PublicationRequestFilter`'s parameters and allowed requests

- Added support for requests without a `status` param

### DIFF
--- a/src/Asset/Asset.js
+++ b/src/Asset/Asset.js
@@ -30,18 +30,17 @@ export class Asset {
   async filter(filters) {
     const { status, type, sort, pagination } = filters.sanitize()
     const tx_status = txUtils.TRANSACTION_STATUS.confirmed
-
     const [assets, total] = await Promise.all([
       db.query(
         SQL`SELECT model.*, row_to_json(pub.*) as publication
-          FROM ${raw(Publication.tableName)} as pub
-          JOIN ${raw(this.tableName)} as model ON model.id = pub.asset_id
-          WHERE status = ${status}
-            AND tx_status = ${tx_status}
-            AND type = ${type}
-            AND ${PublicationQueries.whereisActive()}
-          ORDER BY pub.${raw(sort.by)} ${raw(sort.order)}
-          LIMIT ${raw(pagination.limit)} OFFSET ${raw(pagination.offset)}`
+      FROM ${raw(Publication.tableName)} as pub
+      JOIN ${raw(this.tableName)} as model ON model.id = pub.asset_id
+      WHERE tx_status = ${tx_status}
+        AND type = ${type}
+        AND ${PublicationQueries.whereisActive()}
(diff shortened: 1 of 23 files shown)


src/Asset/Asset.js
```diff
@@ -30,18 +30,17 @@ export class Asset {
...
-            AND type = ${type}
-            AND ${PublicationQueries.whereisActive()}
-          ORDER BY pub.${raw(sort.by)} ${raw(sort.order)}
-          LIMIT ${raw(pagination.limit)} OFFSET ${raw(pagination.offset)}`
+      FROM ${raw(Publication.tableName)} as pub
+      JOIN ${raw(this.tableName)} as model ON model.id = pub.asset_id
+      WHERE tx_status = ${tx_status}
+        AND type = ${type}
+        AND ${PublicationQueries.whereisActive()}
+        ${PublicationQueries.andHasStatus(status)}
```
NicoSantangelo: This is 100% correct, but I think it helps readability and composition if we keep it as `PublicationQueries.hasStatus(status)` so we can use it later in an OR query if we like and we can continue reading left to right here in the query:

```javascript
WHERE tx_status = ${tx_status}
    AND type = ${type}
    AND ${PublicationQueries.whereisActive()}
    AND ${PublicationQueries.hasStatus(status)}
```

what do you think?

src/Publication/Publication.queries.js
```diff
@@ -3,6 +3,8 @@ import { SQL, raw } from '../database'
 
 export const PublicationQueries = Object.freeze({
   whereisActive: () => SQL`expires_at >= EXTRACT(epoch from now()) * 1000`,
+  andHasStatus: status =>
+    status != null ? SQL`AND status = ${status}` : SQL``,
```
NicoSantangelo: If we go with my suggestion of keeping this without the `AND` we can use ` 1 = 1 ` as the default value if status is null

cazala: sabe

src/Asset/Asset.js
cazala: 👍

cazala:
Thanks @NicoSantangelo for the review!

> I'm afraid the change on _TYPE and _STATUS will conflict with https://github.com/decentraland/marketplace/pull/244/files#diff-6e87665f822e8461d3e9dcaf9dd8308cR2. We should be careful while merging

Yup I'll fix all the conflicts in that branch after merging this

> ⚠️ We should remember to update the .env on the front-end once this is deployed!

Will do (:

PR #269, opened 2018-06-21
feat: clear activity log

fmiras:
Create a new button that clears the localstorage completed transactions on activity page

PENDING QA: renewing blockchain to get some txs test data.


webapp/src/components/ActivityPage/ActivityPage.css
```diff
@@ -26,6 +26,10 @@
   margin-bottom: 30px;
 }
 
+.ActivityPage .clear-button {
+  text-align: right !important;
```
abarmat: This button is not aligned to the title that should the number of Activities. It can be fixed with:
```
.ActivityPage .clear-button {
    text-align: right !important;
    float: right;
}
```

abarmat:
empty-page

webapp/src/components/ActivityPage/ActivityPage.css
fmiras: I wanted to avoid using that because aligment of full component is going left like this:

before:
image

after:
image

Still think that's ok?

webapp/src/components/ActivityPage/ActivityPage.js
```diff
@@ -42,10 +41,22 @@ export default class ActivityPage extends React.PureComponent {
...
+      transactionHistory,
+      network,
+      onClear,
+      address
+    } = this.props
 
     return (
       <React.Fragment>
+        <div className="clear-button">
+          <Button size="tiny" className="link" onClick={() => onClear(address)}>
```
NicoSantangelo: We tend to avoid creating new functions on render, more info [here](https://github.com/decentraland/standards/blob/master/standards/react-redux.md#32-avoid-creating-stuff-inside-the-render-function)

webapp/src/modules/transaction/sagas.js
```diff
@@ -66,3 +69,7 @@ function* handleWatchLoadingTransactions(action) {
     }
   }
 }
+
+function* handleClearRequest(action) {
+  yield put(clearTransactionsSuccess(action))
```
NicoSantangelo: Maybe we can just dipatch `getTransactionHistoryFrom(state.data, action.address).map(tx => tx.hash)`, also using the selector `getTransactionHistory(state, address)` here?
cc: @cazala [reducer#83](https://github.com/decentraland/marketplace/pull/269/files#diff-778cbfb33cb84b328e53da374364a2e8R74)

webapp/src/components/ActivityPage/ActivityPage.css
NicoSantangelo: One more comment in this line, the `!important` is not necessary

fmiras:
image
In this case where the button should be?
1) Exactly where it is
2) At pending transactions title level
3) At complete transactions title level

cc: @abarmat @NicoSantangelo @cazala

fmiras:
New update of clear button positions and font sizes:
image

fmiras:
Screenshot of new confirmation prompt: 
image

NicoSantangelo:
@fmiras 
> In this case where the button should be?
>     1. Exactly where it is
>     2. At pending transactions title level
>     3. At complete transactions title level

I vote for ` 1. Exactly where it is`

NicoSantangelo:
@fmiras Minor comments:
* I think making the modal title `Delete transaction history` bigger might ease readability. If the designs look like this, ignore me
* I think the `Activity` title (and subtitle) need a bit of margin bottom to create some space between it and the activities
* Plz rebase 🙏

webapp/src/modules/transaction/sagas.js
cazala: +1 to sending only `address` to `clearTransactionsRequest`

NicoSantangelo:
Ready to merge

PR #271, opened 2018-06-21
feat: change text on mana approved user with no activity

fmiras:
- [x] Implemented
- [x] Tested

### DIFF
--- a/src/Translation/locales/en.json
+++ b/src/Translation/locales/en.json
@@ -1,9 +1,10 @@
 {
   "activity": {
+    "approve": "You can start by approving some MANA on your {settings_link}",
     "completed": "Completed",
     "no_activity": "You have no activity yet.",
     "pending": "Pending",
-    "start": "You can start by approving some MANA on your {settings_link}"
+    "start": "Your able to start trading at the {marketplace}!"
   },
   "address": {
     "its_you": "It's you"
--- a/src/Translation/locales/en.json
+++ b/src/Translation/locales/en.json
@@ -4,7 +4,7 @@
     "completed": "Completed",
     "no_activity": "You have no activity yet.",
     "pending": "Pending",
-    "start": "Your able to start trading at the {marketplace}!"
+    "start": "You are able to start trading at the {marketplace}!"
   },
   "address": {
     "its_you": "It's you"
--- a/webapp/src/components/ActivityPage/ActivityPage.js
+++ b/webapp/src/components/ActivityPage/ActivityPage.js
@@ -13,7 +13,6 @@ import './ActivityPage.css'
 
 export default class ActivityPage extends React.PureComponent {
   static propTypes = {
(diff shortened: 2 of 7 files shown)


src/Translation/locales/en.json
```diff
@@ -1,9 +1,10 @@
 {
   "activity": {
+    "approve": "You can start by approving some MANA on your {settings_link}",
     "completed": "Completed",
     "no_activity": "You have no activity yet.",
     "pending": "Pending",
-    "start": "You can start by approving some MANA on your {settings_link}"
+    "start": "Your able to start trading at the {marketplace}!"
```
eordano: "You are"

fmiras: My bad! fixed 7921b80

NicoSantangelo:
Testing the branch I got this:

image

webapp/src/components/ActivityPage/ActivityPage.js
```diff
@@ -6,16 +6,18 @@ import { locations } from 'locations'
...
 
-import { transactionType } from 'components/types'
+import { transactionType, walletType } from 'components/types'
 import { t, t_html } from 'modules/translation/utils'
 
 import './ActivityPage.css'
 
 export default class ActivityPage extends React.PureComponent {
   static propTypes = {
+    hasAnyPermissionOn: PropTypes.bool.isRequired,
```
NicoSantangelo: Whoops!

webapp/src/components/ActivityPage/ActivityPage.js
```diff
@@ -25,17 +27,41 @@ export default class ActivityPage extends React.PureComponent {
...
       <div className="empty">
         <p>
           {t('activity.no_activity')}
           <br />
-          {t_html('activity.start', {
-            settings_link: (
-              <Link to={locations.settings}>{t('global.settings')}</Link>
-            )
-          })}
+          {this.hasAnyPermissionOn()
```
NicoSantangelo: **Question**:
The wording seems a bit confusing to me, technically you can't _trade_ on the marketplace if you don't have `approvedBalance` and `isLandAuthorized`, but even worse if you only set mortgage permissions.

I'm not sure about this, I'd like @abarmat / @eordano to take a look, it's a minor observation

fmiras: Good observation! I just changed to `hasTradingPermissions` so it only suggest you to move to marketplace if you are able to operate, anyway we should reanalyze how we should behave in case that someone approve mortgage and not the others as @NicoSantangelo says.

abarmat: I think in a way you are allowed to trade once you have both the MANA and LAND permissions regardless of Mortgages, etc.

NicoSantangelo: Superb

PR #272, opened 2018-06-22
chore: update seed.js script to current publication model

fmiras:
Added asset_id, block_number and support both timestamp and ms dates

### DIFF
--- a/scripts/seed.js
+++ b/scripts/seed.js
@@ -75,9 +75,7 @@ function getRandomColumnValue(columnName, tableName) {
     case 'id':
       return undefined
     case 'asset_id':
-      return `${Math.floor(Math.random() * 307) - 153},${Math.floor(
-        Math.random() * 307
-      ) - 153}`
+      return `${getRandomCoordinate()},${getRandomCoordinate()}`
     case 'address':
     case 'owner':
     case 'buyer':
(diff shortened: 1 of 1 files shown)


scripts/seed.js
```diff
@@ -71,62 +71,58 @@ const seed = {
...
     default: {
       if (columnName.includes('price')) {
-        value = faker.random.number(10000000) // 10M max
+        return faker.random.number(10000000) // 10M max
       } else if (columnName.includes('_count')) {
-        value = faker.random.number(1000) // 1k max
+        return faker.random.number(1000) // 1k max
       } else if (columnName.includes('_at')) {
-        value = faker.date.recent()
+        return faker.random.number(1520000000000)
```
NicoSantangelo: Why not `faker.date.recent().getTime()` ? to avoid dealing with a `random`

scripts/seed.js
```diff
@@ -71,62 +71,58 @@ const seed = {
...
 function getRandomColumnValue(columnName, tableName) {
-  let value = null
-
   switch (columnName) {
     case 'id':
-      value = undefined
-      break
+      return undefined
+    case 'asset_id':
+      return `${Math.floor(Math.random() * 307) - 153},${Math.floor(
```
NicoSantangelo: I think that `Math.floor(Math.random() * 307) - 153` might not be completely obvious if the reader doesn't know that asset_id represents.
How about `getRandomCoordinate()` ?

fmiras:
@NicoSantangelo I changed the `value=` approach to a return-based so it's similar as a reducer, where you reach the value you are sure you want to return and you avoid executing more code. Because the value wasn't being muted after the switch statement I thought that it was going to be cleaner, so we save a assignment statement on each case and we just return it. Also, when a function is getting bigger and bigger, having some more points of exits makes easily a programmer to analyse the code because it's ensure that the code exits in that line and not later. I think that is very trivial anyway and is open to a lot of types of thinking, so I you think that the older was better I can just undo this 👍

NicoSantangelo:
@fmiras no need!, but I wanted to know your thoughts on the matter

PR #279, opened 2018-06-26
feat: link addresses on tx history

fmiras:


webapp/src/components/ParcelDetailPage/ParcelDetail/ParcelTransactionHistory/ParcelTransactionHistory.js
```diff
@@ -36,8 +39,10 @@ export default class ParcelTransactionHistory extends React.PureComponent {
   renderAddress(address) {
     return (
       <div className="address-wrapper" title={address}>
-        <AddressBlock address={address} scale={4} hasTooltip={false} />&nbsp;
-        <span className="short-address">{shortenAddress(address)}</span>
+        <Link to={locations.profilePage(address)}>
+          <AddressBlock address={address} scale={4} hasTooltip={false} />&nbsp;
```
NicoSantangelo: Because `AddressBlock` renders a link (`<a>`) React complains that you shouldn't render a link inside of another link.
To avoid this we could pass `hasLink` false to `AddressBlock`

fmiras: Yes! That's my bad, I always forget to check on react warning on console, already fixed on 97526c9!

PR #323, opened 2018-08-08
Add authorization warnings and increase outstanding amount 1 hour

nachomazzara:
This branch has some improvement for mortgages and also is the one we need to use to test the feature


shared/mortgage.js
```diff
@@ -90,6 +91,7 @@ export function getActiveMortgageByBorrower(
  */
 
 export function getMortgageOutstandingAmount(mortgage) {
+  const oneHour = 60 * 60 // used to ensure the total payment of the mortgage
```
NicoSantangelo: If you're already naming the constant `oneHour`, I think it's safe to assign it to `3600` instead of `60 * 60`

shared/mortgage.js
```diff
@@ -22,8 +22,9 @@ export const isMortgageDefaulting = mortgage =>
   new Date().getTime() > parseInt(mortgage.is_due_at * 1000, 10)
 export const isMortgageDefaulted = mortgage =>
   hasStatus(mortgage, MORTGAGE_STATUS.defaulted) ||
-  (!isMortgagePaid(mortgage) &&
-    new Date().getTime() <
+  (!isMortgagePending(mortgage) &&
```
NicoSantangelo: Question: Isn't this check included in `!isMortgagePaid()` ?

shared/mortgage.js
```diff
@@ -169,3 +171,9 @@ export function getMortgageTimeLeft(mortgage) {
 export function getMortgageDefaultedStatus() {
   return MORTGAGE_STATUS.defaulted
 }
+
+export function getMortgageStatus(mortgage) {
```
NicoSantangelo: Isn't this very same logic used on `ParcelMortgage.js`? Shouldn't we refactor here? (delete `getMortgageDefaultedStatus` too)

webapp/src/components/ParcelDetailPage/ParcelDetail/ParcelMortgage/ParcelMortgage.js
```diff
@@ -27,23 +27,13 @@ export default class ParcelMortgage extends React.PureComponent {
 
   render() {
     const { mortgage } = this.props
+
     return (
-      <Grid className={`ParcelMortgageDetail ${mortgage.status}`}>
+      <Grid className={`ParcelMortgageDetail ${getMortgageStatus(mortgage)}`}>
```
NicoSantangelo: Consider caching the `const mortgageStatus = getMortgageStatus(mortgage)` on a variable if it's going to be called > 2 times

webapp/src/modules/wallet/sagas.js
```diff
@@ -107,7 +107,7 @@ function* handleConnectWalletRequest(action = {}) {
       manaTokenContract.balanceOf(address),
       fetchBalance(address),
       manaTokenContract.allowance(address, marketplaceAddress),
-      landRegistryContract.isApprovedForAll(marketplaceAddress, address),
+      landRegistryContract.isApprovedForAll(address, marketplaceAddress),
```
NicoSantangelo: ⚠️ do we want this on master already?
(I think we need to downgrade the contract and re-deploy the new one on another address)

shared/mortgage.js
nachomazzara: What do you mean with `Isn't this very same logic used on ParcelMortgage.js? Shouldn't we refactor here? ` ?

webapp/src/modules/wallet/sagas.js
nachomazzara: I can add a comment to revert that change before marge it to master. What do you think?

PR #338, opened 2018-08-17
feat: transfer estates

### LINKED ISSUE #308: Transfer Estate
call `safeTransfer` from `EstateRegistry`

nachomazzara:
Close #308


webapp/src/components/TransferAssetForm/TransferAssetForm.css
```diff
@@ -0,0 +1,13 @@
+.TransferAssetForm .ui.input.address-input {
+  max-width: 450px !important;
+  display: block;
+}
+
+.TransferAssetForm .ui.input.address-input input {
+  width: 100% !important;
```
NicoSantangelo: Are these `!important`s necessary?
even if they're copied from another .css file I think we can avoid them using specificity

webapp/src/components/EstateDetailPage/EstateDetail.js
```diff
@@ -53,13 +53,16 @@ export default class EstateDetail extends React.PureComponent {
...
+                    onEditMetadata={onEditMetadata}
+                    assetId={estate.asset_id}
+                  />
                 </Grid.Column>
               )}
               {allParcels && (
                 <React.Fragment>
                   <Grid.Column
-                    width={8}
+                    width={isOwner ? 8 : 16}
```
NicoSantangelo: This is kind of weird. Maybe abstracting into a constant? naming floating numbers it's (almost) always a good idea

monitor/reducers/estateReducer.js
```diff
@@ -92,17 +92,17 @@ export async function estateReducer(events, event) {
...
 
       const last_transferred_at = await new BlockTimestampService().getBlockTime(
         block_number
       )
 
       await Estate.update(
-        { owner: to.toLowerCase(), last_transferred_at },
-        { asset_id: estateId }
+        { owner: _to.toLowerCase(), last_transferred_at },
+        { asset_id: _tokenId }
```
NicoSantangelo: Do we have a companion PR on https://github.com/decentraland/land for this?

webapp/src/components/TransferAssetForm/TransferAssetForm.css
nachomazzara: The `max-width` it is being used but we can remove both due other inputs don't have a `max-width`

monitor/reducers/estateReducer.js
nachomazzara: No, `EstateRegistry` inherit from the [ERC721 standard](https://github.com/OpenZeppelin/openzeppelin-solidity/blob/master/contracts/token/ERC721/ERC721Basic.sol#L42) which `transfer` method has that names for `args`

PR #350, opened 2018-08-18
Update ja.json

EijiGard:
some adjustments of Japanese translations


src/Translation/locales/ja.json
```diff
@@ -149,7 +149,7 @@
     "en": "English",
```
abarmat: Can we translate Spanish, French... and the rest of the languages to Japanese too?

EijiGard: I just updated them:) I was wondering if they're written in Japanese, a non-Japanese reader can't read and go back to their language?

PR #379, opened 2018-08-27
show estate names

### LINKED ISSUE #357: Show estate names instead of estate ids in estate activities
Show estate names instead of estate ids in estate activities

cazala:
Closes #357 Shows estate names instead of ids in activity 

screen shot 2018-08-27 at 4 52 24 pm

### DIFF
--- a/webapp/src/components/ActivityPage/Transaction/Transaction.js
+++ b/webapp/src/components/ActivityPage/Transaction/Transaction.js
@@ -64,7 +64,7 @@ export default class Transaction extends React.PureComponent {
   }
 
   handleTransactionClick = e => {
-    if (e.target.nodeName !== 'A') {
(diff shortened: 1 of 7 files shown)


webapp/src/components/ActivityPage/Transaction/Transaction.js
```diff
@@ -270,7 +272,7 @@ export default class Transaction extends React.PureComponent {
       case CREATE_ESTATE_SUCCESS: {
         const { estate } = payload
         return t_html('transaction.create_estate', {
-          parcels_link: this.renderParcelsLink(estate.data.parcels)
+          name: estate.data.name
```
nachomazzara: I think the method `renderParcelsLink` is not using anymore, we can delete it

cazala: ✅ done (faf674f6f69bb2f92c877179f8e50e5396d7efbf)

PR #389, opened 2018-08-28
fix: handle empty estate

### LINKED ISSUE #384: Add a message when you visit an empty Estate

cazala:
Closes #384 

Currently empty (aka deleted) estates show a broken page (only navbar and footer, and footer is _en cualquiera_)

This PR adds a screen for empty estates:

screen shot 2018-08-28 at 4 15 48 pm

### DIFF
--- a/webapp/src/components/EstateDetailPage/EstateDetailPage.js
+++ b/webapp/src/components/EstateDetailPage/EstateDetailPage.js
@@ -41,8 +41,7 @@ export default class EstateDetailPage extends React.PureComponent {
     return (
       <Estate assetId={assetId} x={x} y={y}>
         {(estate, isOwner, wallet) =>
-          estate.data.parcels.length > 0 &&
-          (isNewAsset(estate) || isEditing) ? (
+          isNewAsset(estate) || isEditing ? (
             <EditEstate
               estate={estate}
               isCreation={isNewAsset(estate)}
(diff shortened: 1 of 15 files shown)


NicoSantangelo:
The changeset looks 💯 but I'd like to revise the wording we use for this action. I know I raised this concern before...but I'm annoying like that.

Given that supporting empty states is on the roadmap https://github.com/decentraland/marketplace/projects/4 I think that we can change `delete` for something like `dissolve`

webapp/src/components/EstateDetailPage/EstateDetailPage.js
```diff
@@ -41,27 +41,26 @@ export default class EstateDetailPage extends React.PureComponent {
...
-                allParcels={allParcels}
-                estate={estate}
-                isOwner={isOwner}
-                onViewAssetClick={onViewAssetClick}
-                onEditParcels={onEditParcels}
-                onEditMetadata={onEditMetadata}
-              />
-            )
-          ) : null
+          estate.data.parcels.length > 0 &&
```
nachomazzara: I think we can remove `estate.data.parcels.length > 0` I added it to not render anything in case the estate is empty. As we have that check [here](https://github.com/decentraland/marketplace/pull/389/files#diff-dc94e71ea3fc145084f572f535ae7185R50) maybe it is obsolete to have it here too

PR #395, opened 2018-08-29
feat: show tags at estate card

nachomazzara:


webapp/src/components/ProfilePage/ProfilePage.js
```diff
@@ -114,11 +115,28 @@ export default class ProfilePage extends React.PureComponent {
...
         return (
           <Card.Group stackable={true}>
-            {grid.map(estate => (
-              <EstateCard key={estate.asset_id} estate={estate} />
-            ))}
+            {grid.map(estate => {
+              const estateParcels = estate.data.parcels.reduce(
+                (acc, parcel) => {
+                  const parcelId = buildCoordinate(parcel.x, parcel.y)
+                  if (parcels[parcelId]) {
```
cazala: `parcels` is an array, not an object, so `parcels[parcelId]` will always be `undefined`

could we connect the `parcels` object in the `EstateCard.container.ts` so the parent component doesn't need to have it in order to use it?

PR #431, opened 2018-09-03
fix: check name/descriptions correctly

### LINKED ISSUE #400: Estate with no name when editing should not be allowed

NicoSantangelo:
plz testerino 🙏 
Closes #400

### DIFF
--- a/shared/asset.js
+++ b/shared/asset.js
@@ -184,12 +184,12 @@ export function getType(asset, publications, wallet) {
   return TYPES.taken
 }
 
-export function isValidName(name) {
-  return name <= 50
+export function isValidName(name = '') {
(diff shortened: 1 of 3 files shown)


shared/asset.js
```diff
@@ -184,12 +184,12 @@ export function getType(asset, publications, wallet) {
   return TYPES.taken
 }
 
-export function isValidName(name) {
-  return name <= 50
+export function isValidName(name = '') {
+  return name.length > 0 && name.length <= 50
```
nachomazzara: This could be `name.length && name.length <= 50`

cazala: TypeScript doesn't think the same :P although javascript allows stuff like that let's stick to consistent return values (in this case, always return a boolean)

PR #440, opened 2018-09-04
fix: use all parcels to determine connections

NicoSantangelo:
@cazala ¯\\_(ツ)_/¯

### DIFF
--- a/shared/parcel.js
+++ b/shared/parcel.js
@@ -20,21 +20,14 @@ export function isParcel(asset) {
   )
 }
 
-export function toParcelObject(
-  parcelsArray,
-  prevParcels = [],
-  normalize = true,
-  redoConnections = false
-) {
+export function toParcelObject(parcelsArray, prevParcels) {
   return connectParcels(
     parcelsArray,
     parcelsArray.reduce((map, parcel) => {
-      map[parcel.id] = normalize
-        ? normalizeParcel(parcel, prevParcels[parcel.id])
-        : parcel
+      map[parcel.id] = normalizeParcel(parcel, prevParcels[parcel.id])
       return map
     }, {}),
-    redoConnections
+    prevParcels
   )
 }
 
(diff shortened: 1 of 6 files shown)


shared/parcel.js
```diff
@@ -20,21 +20,14 @@ export function isParcel(asset) {
   )
 }
 
-export function toParcelObject(
-  parcelsArray,
-  prevParcels = [],
-  normalize = true,
-  redoConnections = false
```
nachomazzara: If we remove the `redoConnections` we will lose the _feature_ of a confirmed tx of removing a parcel from an Estate

cazala: That's taken care by this line: https://github.com/decentraland/marketplace/pull/440/files#diff-fa45755e5e2585857eb74278da106dd6R51

shared/parcel.js
```diff
@@ -21,14 +21,11 @@ export function isParcel(asset) {
...
 export function toParcelObject(parcelsArray, prevParcels) {
-  return connectParcels(
-    parcelsArray,
-    parcelsArray.reduce((map, parcel) => {
-      map[parcel.id] = normalizeParcel(parcel, prevParcels[parcel.id])
-      return map
-    }, {}),
-    prevParcels
-  )
+  const parcelObject = {}
```
nachomazzara: Is it wrong if we call it `parcelsObject`?

shared/parcel.js
nachomazzara: Yep, with the new commit is fixed 👍

shared/parcel.js
cazala: the function is called `toParcelObject` that's why i used `parcelObject`. But for consistency I think we could rename `parcelsArray` to `parcelArray`, right?

nachomazzara: I think that would be better to rename the function to `toParcelsObject` its return an object with all the parcels as an object. And also use `parcelsObject` and what you say `parcelArray`

Clean version:

`toParcelObject` to `toParcelsObject`
`parcelObject` to `parcelsObject`
`parcelsArray` to `parcelArray`

PR #444, opened 2018-09-05
fix: use anchors for external links instead of react-router <Link>

### LINKED ISSUE #439: `Link` adds a `/` at the beginning of the URL breaking blocknumber (on transaction list) and etherscan (on the authorize transaction in the activity page) links

cazala:
Closes #439


NicoSantangelo:
:+1

PR #448, opened 2018-09-05
fix: pending tx estates

### LINKED ISSUE #407: Transaction pending indicator when TX intent in non-related parcel

nachomazzara:
fix #407 

# Showing Estate pending transactions as: 

screen shot 2018-09-05 at 7 23 21 pm

Also, fix the styles of using `Container` instead of `div` on `forms` pages

### Double background color
screen shot 2018-09-05 at 7 19 56 pm

### Not full width on mobile when showing pending tx
screen shot 2018-09-05 at 7 20 18 pm

# Rename some `token_id` to `id` 
I kept critics renames to be approached in other PR (when we decide to remove `token_id` from the db)


webapp/src/modules/estates/actions.js
```diff
@@ -138,7 +150,11 @@ export function deleteEstateSuccess(txHash, estate) {
   return {
     type: DELETE_ESTATE_SUCCESS,
     ...buildTransactionAction(txHash, {
-      estate,
+      estate: {
+        id: estate.id,
+        data: { name: estate.data.name, parcels: estate.data.parcels }
```
NicoSantangelo: Maybe we could have a utils fn to map the parcels leaving only the necessary props `(x,y)` ?

webapp/src/modules/estates/sagas.js
```diff
@@ -185,14 +185,15 @@ function* handleTransferRequest({ estate, to }) {
...
-      contract.safeTransferFrom(oldOwner, to, estate.token_id)
+      contract.safeTransferFrom(oldOwner, to, estate.id)
     )
 
     const transfer = {
       txHash,
       oldOwner,
       to,
-      estate
+      estate: { id: estate.id, data: { name: estate.data.name } },
```
NicoSantangelo: same as estates/actions.js

nachomazzara: Sorry, I don't get it. What do you mean? We are not passing `parcels` here.

webapp/src/modules/estates/actions.js
nachomazzara: Parcels have an object `{ x, y }`  here

webapp/src/components/EstateDetailPage/EditEstateMetadata/EstateModal/EstateModal.js
```diff
@@ -75,40 +76,42 @@ export default class EstateModal extends React.PureComponent {
...
-            <React.Fragment>
-              <div className="modal-children">{children}</div>
-              <br />
-            </React.Fragment>
-          ) : null}
-          {hasCustomFooter ? null : (
-            <div>
-              <Grid.Column>
-                <div className="modal-buttons">
+          <div className="modal-children">
```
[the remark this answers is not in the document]
nachomazzara: resolved offline

webapp/src/modules/estates/sagas.js
nachomazzara: Added comments as we talked offline

PR #467, opened 2018-09-12
feat: estate sanity check and abstractions

### LINKED ISSUE #347: Update sanity-check, self-heal and scripts if needed to handle the new marketplace and estates

NicoSantangelo:
`npm run sanity-check -- # --skip Parcel,Publication,Estate --self-heal`


sanity/SanityActions.js
```diff
@@ -0,0 +1,81 @@
...
+import { MonitorActions } from '../monitor/MonitorActions'
+import { main as indexMissingEvents } from '../monitor/program'
+import { doctors } from './doctors'
+
+const log = new Log('Sanity')
+
+export class SanityActions {
+  async run(options) {
+    const validations = this.getValidations(options.skip)
+    const diagnostics = []
```
cazala: the plural of `diagnosis` is `diagnoses` 🤓

abarmat:
Gathering a dump and log for some inconsistencies that the script is not fixing.

sanity/SanityActions.js
NicoSantangelo: <3

PR #471, opened 2018-09-12
feat: migrate ls v2

nachomazzara:
Remove `receipt` from old `transactions` and save `receipt` on demand. In this PR for `createEstate` action


abarmat:
Receipts are now null instead of the key removed from the object, is that expected behavior? if that's right I'm ok with the PR.

nachomazzara:
> Receipts are now null instead of the key removed from the object, is that expected behavior? if that's right I'm ok with the PR.

For new estate creations, we should read from `receipt` so, I think it is ok to have set it to null

PR #484, opened 2018-09-25
[WIP] feat: mkt v2 authorization

NicoSantangelo:


webapp/src/components/SettingsPage/SettingsPage.js
```diff
@@ -46,72 +38,54 @@ export default class SettingsPage extends React.PureComponent {
...
+  handleTokenApproval = (checked, contractName, tokenContractName) => {
+    this.props.onApproveToken(checked, contractName, tokenContractName)
   }
 
-  getApproveTransaction() {
-    // Transactions are ordered, the last one corresponds to the last sent
-    const { approveManaTransactions } = this.props.wallet
-    return approveManaTransactions[approveManaTransactions.length - 1]
-  }
+  getPendingTransactions() {
```
NicoSantangelo: Should I leave this here or on the container?

webapp/src/components/BuyParcelByMortgagePage/BuyParcelByMortgagePage.container.js
```diff
@@ -2,26 +2,39 @@ import { withRouter } from 'react-router'
...
 
 const mapState = (state, ownProps) => {
   const { x, y } = getMatchParamsCoordinates(ownProps)
+  const wallet = getWallet(state)
+
+  let authorization
+
+  if (wallet) {
+    authorization = getAuthorizations(state)[wallet.address]
+  }
```
NicoSantangelo: this pattern:

```javascript
  let authorization
   if (wallet) {
    authorization = getAuthorizations(state)[wallet.address]
  }
```

repeats on every container that needs `authorization`. Do you thinks is better to do something like `const authorizations = getAuthorizations(state)[getAddress(state)]` ?

webapp/src/modules/authorization/actions.js
```diff
@@ -0,0 +1,145 @@
+import { buildTransactionAction } from 'modules/transaction/utils'
+
+// Fetch authorization
+
+export const FETCH_AUTHORIZATION_REQUEST = '[Request] Fetch Authorization'
+export const FETCH_AUTHORIZATION_SUCCESS = '[Success] Fetch Authorization'
+export const FETCH_AUTHORIZATION_FAILURE = '[Failure] Fetch Authorization'
```
NicoSantangelo: Do you think this should be pluralized?

We're fetching multiple authorizations for a user, but the module is called authorization and contains multiple instances of the same authorization instance for each address.

I'm not sure if using `authorization.allowances, authorization.approvals` is better than using `authorizations.allowances, authorizations.approvals` everywhere

nachomazzara: We have `transaction` for transactions and `parcels` for parcel. As we should need to be consistent when naming, I prefer plural for `authorizations` instead of `authorization`. 

The actions are ok if you are always fetching only one authorization

webapp/src/components/BuyParcelByMortgagePage/BuyParcelByMortgagePage.container.js
nachomazzara: What do you think of using a HOC which handles authorizations for children components?

```
class WithAuthorization extends React.Component {}
withAuthorization(connect(toState, toDispatch)(BuyParcelByMortgagePage))
```

Maybe we can do it in another PR

webapp/src/modules/address/selectors.js
```diff
@@ -33,30 +43,33 @@ export const getAddresses = createSelector(
...
+          mortgagedParcels,
           parcels,
           parcelsById,
-          estates: estates.filter(e => e.owner === address),
+          estates,
           estatesById,
           contributions,
-          publishedParcels
+          publishedParcels,
+          authorization
```
NicoSantangelo: I added this here for completeness sake, but I'm not using it. @cazala do you think is better to keep it here anyways?

webapp/src/modules/publication/utils.js
```diff
@@ -0,0 +1,7 @@
+import { getContractAddress } from 'modules/wallet/utils'
+
+export function isLegacyPublication(publication) {
```
nachomazzara: Is it worth to put in share instead? Is there any context where we should use it server-side?

webapp/src/components/SettingsPage/SettingsPage.js
```diff
@@ -46,72 +38,54 @@ export default class SettingsPage extends React.PureComponent {
...
+    const { allowTransactions, approveTransactions } = this.props
+    const isPending = transaction =>
+      transaction.status === txUtils.TRANSACTION_STATUS.pending
```
cazala: you should use `isPending(transaction.status)` from `@dapps/modules/transaction/utils` which handles also `null` and `queued` transaction status (which are "pending")

webapp/src/modules/publication/utils.js
NicoSantangelo: Yes! but the problem was that I can't use getContractAddress, as it's using the `eth` singleton and that's not available on `shared/`
I've decided to leave it on publication/utils because it's "temporal" code anyways, it'll disappear with the old publications

webapp/src/modules/authorization/actions.js
NicoSantangelo: The thing is, the module can be singular because is the _name_ of the module (I think we're using those and being consistent on the other dapps) @cazala back me up if I'm not correct

The problem I see here with `FETCH_AUTHORIZATION_REQUEST` is that one authorization consists of several things, multiple allowances and approvals, so it might be a confusing action name

webapp/src/components/BuyParcelByMortgagePage/BuyParcelByMortgagePage.container.js
NicoSantangelo: I like that! I should put more thought into it but I like that it's composable and easily removable, but yes we can leave it for another PR

webapp/src/modules/authorization/actions.js
cazala: back in the day we decided to use singular names for modules because it's more aligned with DDD: it's the name of the module, ie "the transaction module", "the parcel domain". But we  screwed up and let a few plurals slip in on this dapp /shrug.

webapp/src/modules/publication/utils.js
nachomazzara: ok, cool! anways if we need it also server-side. We can refactor it sending the address as an input parameter

webapp/src/modules/wallet/selectors.js
```diff
@@ -26,30 +21,9 @@ export const isConnecting = state =>
...
-  (
-    wallet,
-    addresses,
-    approveManaTransactions,
-    authorizeLandTransactions,
-    approveMortgageForManaTransactions,
-    approveMortgageForRCNTransactions
-  ) => {
+  getData,
+  getAddresses,
```
cazala: bad rebase? ☝️

cazala:
```
state => getData(state),
state => getAddresses(state),	  
getData,
getAddresses,
```

webapp/src/modules/address/selectors.js
cazala: 👍

webapp/src/modules/authorization/actions.js
cazala: lets leave it like this 👍

webapp/src/components/SettingsPage/SettingsPage.js
cazala: if this component only uses the pending transactions let's filter them in the container (we could use a selector for it) and pass only those down to this component

PR #489, opened 2018-09-28
feat: deposit required & fix location

nachomazzara:
# Mortgage deposit

- Add deposit and check if user balance is greater than it

screen shot 2018-09-27 at 10 50 40 pm

- Fix pay mortgage location

- If expires is greater than 10years set that as timestamp


webapp/src/components/BuyParcelByMortgagePage/BuyParcelByMortgagePage.js
```diff
@@ -80,7 +80,11 @@ export default class BuyParcelByMortgagePage extends React.PureComponent {
       return this.renderNotConnected()
     }
 
-    const { isMortgageApprovedForMana, isMortgageApprovedForRCN } = wallet
+    const {
+      isMortgageApprovedForMana,
+      isMortgageApprovedForRCN,
```
cazala: i think this doesn't work with the current authorizations model

cc @NicoSantangelo

NicoSantangelo: Yup! it was bound to be a conflict post merge :(

PR #491, opened 2018-09-29
feat: add updateOperator

eordano:


cazala:
looks like the linter is complaining ☝️

NicoSantangelo:
Two things:
- To add a new validation to the sanity check, you should go to https://github.com/decentraland/marketplace/blob/master/sanity/doctors/ParcelDoctor.js#L45. The `ParcelDiagnosis` will take care of the fixing part in this case, so no code needed there, if you want, you could add an `update` removing `update_operator` [here](https://github.com/decentraland/marketplace/blob/master/sanity/doctors/ParcelDoctor.js#L76), but I think that if the blockchain events are re-fetched it should work fine.
- The tests are failing just because we need to add `update_operator` to the expected object

migrations/1538238012345_parcels-add-update-operator.js
```diff
@@ -0,0 +1,23 @@
+import { Parcel } from '../src/Parcel'
+import { Estate } from '../src/Estate'
```
cazala: The above two come from `../src/Asset` (as of #492)

cazala:
@eordano looking good but need rebase

PR #492, opened 2018-10-01
feat: estates api

### LINKED ISSUE #341: API endpoint for estates `/estates` which mimics the current `/parcels`
Also add `/marketplace` endpoint with `asset_type` as query param

NicoSantangelo:
Closes #341 

```
/marketplace
status required
filter by publication
asset_type not required

/parcels and /states
status not required
if (!status) return everything

/assets
same as /parcels, /estates, etc, but with explicit asset_type
asset_type required
```


src/Publication/Publication.service.js
```diff
@@ -1,23 +1,27 @@
...
     }
 
-    return {
-      [ASSET_TYPES.parcel]: this.Parcel,
-      [ASSET_TYPES.estate]: this.Estate
-    }[assetType]
+    return ASSETS[assetType]
+  }
+
+  getPublicableAssets(assetType) {
```
cazala: it looks like the `assetType` argument is not used in this function

src/Publication/Publication.service.js
```diff
@@ -1,23 +1,27 @@
...
+    const publicableAssets = {}
+
+    for (const key in PUBLICATION_ASSET_TYPES) {
+      publicableAssets[key] = ASSETS[key]
+    }
+
+    return publicableAssets
```
cazala: This return value is an `object` but it's expected to be an `array` [here](https://github.com/decentraland/marketplace/pull/492/files#diff-8b7ae6d8fe218adc457ca3f6733286c1R13) and [there](https://github.com/decentraland/marketplace/pull/492/files#diff-8b7ae6d8fe218adc457ca3f6733286c1R17)

src/server.js
```diff
@@ -44,14 +44,16 @@ if (env.isDevelopment()) {
 const router = new express.Router()
 app.use('/v1', router)
 
-new TranslationRouter(router).mount()
-new PublicationRouter(router).mount()
-new MortgageRouter(router).mount()
+new AssetRouter(router).mount()
+new ParcelRouter(router).mount()
```
cazala: `ParcelRouter` has to be the last one to be mounted... otherwise the `parcels/:x/:y` endpoint handles the requests for `/parcels/:address/mortgages`.

ie:

Request:

```
GET /parcels/0x0de1ec708665b93478bd0ed264e60b68a8663bb4/mortgages
```

Response:

```json
{
    "ok": false,
    "data": {},
    "error": "Invalid coordinate \"y\" must be an integer"
}
```

Although i think we should change the `/parcels/:address/mortgages` to be `/addresses/:address/mortgages` since the uri param is an identifier for the `addresses` resource, not `parcels`.

NicoSantangelo: I'll do both!

src/Publication/Publication.service.js
NicoSantangelo: good catch! didn't test my last changes correctly

src/Publication/Publication.service.js
NicoSantangelo: 👍

cazala:
👍

PR #511, opened 2018-10-04
fix: estate publications on map

cazala:
This PR fixes a bug in the branch `fix/list-estates` that causes the publications of Estates fetched by the `FETCH_MAP_REQUEST` to be left out of the app state. This resulted in Estate publications not showing up in the Atlas or in the EstateDetail page.

### DIFF
--- a/webapp/src/lib/webworker/WebWorkerOnMessage.js
+++ b/webapp/src/lib/webworker/WebWorkerOnMessage.js
@@ -5,7 +5,20 @@ export function WebWorkerOnMessage(event) {
   let result = {}
 
   switch (action.type) {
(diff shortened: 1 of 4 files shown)


webapp/src/lib/webworker/WebWorkerOnMessage.js
```diff
@@ -17,7 +30,7 @@ export function WebWorkerOnMessage(event) {
       break
     }
     case 'FETCH_ADDRESS_ESTATES_REQUEST': {
-      const { estates, allEstates } = action
+      const { estates } = action
```
nachomazzara: 👍

PR #527, opened 2018-10-07
feat: request estate data when navigate

### LINKED ISSUE #519: When switching between estate details, the publications of the new estate selecteds are not fetched
When switching between estate details, the publications of the new estate selecteds are not fetched

nachomazzara:
Close #519


cazala:
we have this same logic in the `components/Parcel` component, what if we move it to the `components/Asset` component, since it's something that we want for any asset i guess (and keeps the code dry)

PR #549, opened 2018-10-09
fix: switching between parcels and estate details

cazala:
This PR combines the `EstateDetailPage` and `ParcelDetailPage` under a parent `AssetDetailPage`. This removes some duplicated logic, and most importantly it prevents the `AssetPreviewHeader` (fka AssetDetailPage) to remount when switching between parcel and estate detail pages, preserving the zoom and avoiding the flickering.

I also took this change to split the EstateEditPage from the EstateDetailPage, since those are very different components with their own logic and purpose.


webapp/src/components/AssetDetailPage/AssetDetailPage.container.js
```diff
@@ -1,25 +1,25 @@
...
 import AssetDetailPage from './AssetDetailPage'
+import { ASSET_TYPES } from '../../shared/asset'
+import { buildCoordinate } from '../../shared/parcel'
 
-const mapState = (state, { asset }) => {
-  const { x, y } = getCenterCoords(asset)
+const mapState = (state, { location, match }) => {
+  const assetType = Object.keys(ASSET_TYPES)
+    .map(key => ASSET_TYPES[key])
+    .find(assetType => location.pathname.includes(ASSET_TYPES[assetType]))
```
NicoSantangelo: Support for old parcel URLs breaks!

image

shared/parcel.js
```diff
@@ -20,6 +20,10 @@ export function isParcel(asset) {
   )
 }
 
+export function isParcelId(id = '') {
+  return /^-?\d+,-?\d+$/.test(id)
```
NicoSantangelo: This makes me weary. We can have a new asset down the line that uses similar ids and the check will break.

To avoid this and the comment here https://github.com/decentraland/marketplace/pull/549/files#diff-59adc9d1d5a72a75a482f79a544fef9dR13 what do you think about adding the asset_type to the Routes?
Like this:

```jsx
<Route
  exact
  path={this.addLegacySupport(locations.buyParcel())}
  component={withAssetType(BuyParcelPage, ASSET_TYPE.parcel)}
/>
<Route
  exact
  path={this.addLegacySupport(locations.cancelSaleParcel())}
  component={withAssetType(CancelSaleEstatePage, ASSET_TYPE.parcel)}
/>
```

```js
function withAssetType(component, assetType) {
  return props => <Component assetType={assetType} {...props} />
}
```

or we can take a different approache, like using a `<RouteWithAssetType />`

```jsx
<RouteWithAssetType
  exact
  path={this.addLegacySupport(locations.cancelSaleParcel())}
  component={CancelSaleEstatePage}
  assetType=ASSET_TYPE.parcel}
/>
```

whatever you think is best

_code written on GitHub, use at your own risk_

webapp/src/modules/mortgage/sagas.js
```diff
@@ -230,5 +232,8 @@ function* handleClaimMortgageResolutionRequest(action) {
 }
 
 function* handleFetchParcelSuccess(action) {
-  yield put(fetchActiveParcelMortgagesRequest(action.x, action.y))
+  const loadingState = yield select(state => getLoading(state))
+  if (!isLoadingType(loadingState, FETCH_ACTIVE_PARCEL_MORTGAGES_REQUEST)) {
```
nachomazzara: Can we use the `take` instead of `takeLatest` effect to replace this? 

```js
 yield take(
      FETCH_ACTIVE_PARCEL_MORTGAGES_REQUEST,
      handleFetchActiveParcelMortgagesRequest
    )
```

shared/parcel.js
cazala: I used the `render` method of the `Route` component, because doing

```
function withAssetType(component, assetType) {
  return props => <Component assetType={assetType} {...props} />
}
```

that creates a new component on each render, which makes it to remount, bringing back the flickering and causing the internal state to be lost.

webapp/src/components/Parcel/Parcel.js
```diff
@@ -1,28 +1,20 @@
...
+    y: PropTypes.oneOfType([PropTypes.string, PropTypes.number])
   }
 
   render() {
-    const { parcel } = this.props
-    return (
-      <Asset value={parcel} isConnected={this.isConnected} {...this.props} />
-    )
+    let { id, x, y } = this.props
+    if (id == null && !isNaN(x) && !isNaN(y)) {
```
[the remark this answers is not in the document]
cazala: done

webapp/src/modules/mortgage/sagas.js
cazala: done

webapp/src/components/AssetDetailPage/AssetDetailPage.container.js
cazala: fixed in https://github.com/decentraland/marketplace/pull/549/commits/ec4510dc29ac78b6c0cf4d602d697628033c5b39

cazala:
Thanks for the review @NicoSantangelo and @nachomazzara! 
I addressed the requested changes and rebased to master.

webapp/src/components/AssetLoader/AssetLoader.container.js
```diff
@@ -0,0 +1,47 @@
...
+
+  return {
+    isLoading,
+    asset
+  }
+}
+
+const mapDispatch = (dispatch, { id, assetType }) => {
+  const isParcel = assetType === ASSET_TYPES.parcel
+  const [x, y] = isParcel ? splitCoordinate(id) : [null, null]
```
NicoSantangelo: I think this could be `const [x, y] = splitCoordinate(id)` because will return `[0, 0]` if id is not defined and you're already checking for `isParcel` below.

if we want to be extra sure we could add a toString() to splitCoordinate:


```diff
export function splitCoordinate(id) {
  let ids = [0, 0]
  if (id) {
-    ids = id.split(',').map(coord => parseInt(coord, 10))
+    ids = id.toString().split(',').map(coord => parseInt(coord, 10))
  }
  return ids
}
```

PR #584, opened 2018-10-16
feat: estates update operator

### LINKED ISSUE #474: Update Operator permissions for Estates

nachomazzara:
Close #474 

Pending decentraland/land#115


webapp/src/modules/wallet/utils.js
```diff
@@ -100,19 +102,35 @@ export async function fetchBalance(address) {
...
+
+async function getEstateUpdateOperator(tokenId) {
+  const contract = eth.getContract('EstateRegistry')
+  const address = await contract.updateOperator(tokenId)
+  if (eth.utils.isValidAddress(address) && !Contract.isEmptyAddress(address)) {
+    return address
+  }
+
+  return null
+}
```
NicoSantangelo: I don't like having `eth.utils.isValidAddress(address) && !Contract.isEmptyAddress(address)` duplicated. We're not exporting `getParcelUpdateOperator` or `getEstateUpdateOperator`, maybe we can do something like returning the address here and making the check once on `getAssetUpdateOperator`?

webapp/src/components/ActivityPage/Transaction/Transaction.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import { Link } from 'react-router-dom'
-import { Segment, Grid } from 'semantic-ui-react'
+import { Segment, Grid } from 'semantic-ui-react/dist/commonjs'
```
NicoSantangelo: Whoops the original was correct

webapp/src/components/EstateDetailPage/EstateDetailPage.js
```diff
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types'
 
 import Estate from 'components/Estate'
 import { parcelType, publicationType } from 'components/types'
-import EstateDetail from './EstateDetail'
+import EstateDetail from 'components/EstateDetailPage/EstateDetail'
```
NicoSantangelo: We tend to keep the relative imports for components on the same folder, but you don't really need to change this

webapp/src/components/ActivityPage/Transaction/Transaction.js
nachomazzara: God 🤒

webapp/src/components/EstateDetailPage/EstateDetailPage.js
nachomazzara: God 🤒

nachomazzara:
@NicoSantangelo which kind of repetition? could you explain it a little bit

NicoSantangelo:
Sure!

I think the two things that stuck out to me were:
- [handleManageEstateRequest](https://github.com/decentraland/marketplace/pull/584/files#diff-a6d8d74995485608dd46d2142f5a62bfR215) vs handleParcelEstateRequest
- ManageEstatePage and ManageParcelPage

I know abstracting the sagas can be a bit of a pain, because you don't have a management module to do something like `FETCH_ASSET_PUBLICATIONS_REQUEST` (which is the same action for every asset). And abstracting the components is a pain.

But the thing that made me weary is that if a new asset appears we'll need to repeat quite a bit of code and that looks like future tech debt

nachomazzara:
> Sure!
> 
> I think the two things that stuck out to me were:
> 
> * [handleManageEstateRequest](https://github.com/decentraland/marketplace/pull/584/files#diff-a6d8d74995485608dd46d2142f5a62bfR215) vs handleParcelEstateRequest
> * ManageEstatePage and ManageParcelPage
> 
> I know abstracting the sagas can be a bit of a pain, because you don't have a management module to do something like `FETCH_ASSET_PUBLICATIONS_REQUEST` (which is the same action for every asset). And abstracting the components is a pain.
> 
> But the thing that made me weary is that if a new asset appears we'll need to repeat quite a bit of code and that looks like future tech debt

I like the idea to have a management module. I will go with that one :)

NicoSantangelo:
I don't see update operators on estates! I always see

image

webapp/src/modules/analytics/track.js
```diff
@@ -104,12 +105,15 @@ add(
...
 
-add(MANAGE_PARCEL_SUCCESS, 'Manage LAND Permissions', action => ({
-  x: action.parcel.x,
-  y: action.parcel.y,
-  address: action.address,
-  revoked: action.revoked
-}))
+add(
+  MANAGE_ASSET_SUCCESS,
+  action => addAssetType('Manage Asset Permissions', action.asset_type),
```
cazala: This will end up as `Manage Asset Permissions Parcel` and `Manage Asset Permissions Estate`, doesn't sound better to just be called something like `Manage Parcel` and `Manage Estate` ?

PR #587, opened 2018-10-17
feat: add error message when length is exceeded

### LINKED ISSUE #450: Add error description when name or description length is more than allowed
This also means creating a generic way to handle and display errors in forms.

nachomazzara:
Close #450 
screen shot 2018-10-17 at 3 35 32 pm

### DIFF
--- a/webapp/src/components/CreateEstatePage/EditEstate/EditEstateForm/EditEstateForm.css
+++ b/webapp/src/components/CreateEstatePage/EditEstate/EditEstateForm/EditEstateForm.css
@@ -1,1 +1,0 @@
-.EditEstateForm {}
--- a/webapp/src/components/EditEstatePage/EditEstateMetadata/EditEstateMetadataForm/EditEstateMetadataForm.js
+++ b/webapp/src/components/EditEstatePage/EditEstateMetadata/EditEstateMetadataForm/EditEstateMetadataForm.js
@@ -131,6 +131,7 @@ export default class EditEstateMetadataForm extends React.PureComponent {
             type="text"
             value={name}
             onChange={this.handleNameChange}
+            error={!this.isValidName(name)}
             autoFocus
           />
         </Form.Field>
@@ -140,6 +141,7 @@ export default class EditEstateMetadataForm extends React.PureComponent {
             type="text"
             value={description}
             onChange={this.handleDescriptionChange}
+            error={!this.isValidDescription(description)}
           />
         </Form.Field>
         <TxStatus.Idle isIdle={isTxIdle} />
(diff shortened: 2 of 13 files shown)


nachomazzara:
I know EditParcelform and EditEstateMetadata are similar. What do you think of creating a technical debt issue to have one? @cazala @NicoSantangelo

NicoSantangelo:
Some feedback:

### 1

I don't like that we lost the indication of an error on the input itself

image

It allows the user to know somethings wrong before pressing submit

### 2

The error message for empty names is really confusing
image

-----

I'm not really crazy about bit error messages.
I'd prefer having something like twitters counter, where you see how many characters you have left and you can get away with only showing the error on the input as we did before. I could be on the right, where the warning for the content link for parcels currently is.
It's 100% debatable

webapp/src/components/CreateEstatePage/EditEstate/EditEstateForm/EditEstateForm.css
```diff
@@ -1 +0,0 @@
-.EditEstateForm {}
```
NicoSantangelo: We can maybe delete this file?

webapp/src/components/EditEstatePage/EditEstateMetadata/EditEstateMetadataForm/EditEstateMetadataForm.js
```diff
@@ -44,6 +50,7 @@ export default class EditEstateMetadataForm extends React.PureComponent {
         name: event.target.value
       }
     })
+    this.setState({ formErrors: [] })
```
NicoSantangelo: Is it necessary to update the state of an unmounted component?

webapp/src/components/EditEstatePage/EditEstatePage.js
```diff
@@ -10,8 +10,8 @@ import { getInitialEstate } from 'shared/estate'
 export default class EditEstatePage extends React.PureComponent {
   static propTypes = {
     id: PropTypes.string,
-    x: PropTypes.string,
-    y: PropTypes.string,
+    x: PropTypes.number,
+    y: PropTypes.number,
```
NicoSantangelo: What changed here?

nachomazzara: React show warning cause should expect number instead of string

webapp/src/components/CreateEstatePage/EditEstate/EditEstateForm/EditEstateForm.css
nachomazzara: It is deleted 🤷‍♂️

webapp/src/components/EditEstatePage/EditEstateMetadata/EditEstateMetadataForm/EditEstateMetadataForm.js
nachomazzara: We need to clean the errors if the user change the fields

cazala:
can we add a ```.ui.message > .close.icon {
  margin: 5px;
}``` so the X is aligned?

PR #588, opened 2018-10-17
feat: new mkt cli

### LINKED ISSUE #506: Update `mktcli.js` to support the new marketplace

NicoSantangelo:
Closes #506

### DIFF
--- a/scripts/mktcli.js
+++ b/scripts/mktcli.js
@@ -64,20 +64,10 @@ const main = {
       .action(
         asSafeAction(async (assetId, assetType) => {
           const asset = await getAssetFromCLIArgs(assetId, assetType)
-          let contract
 
-          switch (assetType) {
-            case ASSET_TYPES.parcel: {
-              contract = eth.getContract('LANDRegistry')
-              break
-            }
-            case ASSET_TYPES.estate: {
-              contract = eth.getContract('EstateRegistry')
-              break
-            }
-          }
-
+          const contract = getContractByAssetType(assetType)
           const owner = await contract.ownerOf(asset.token_id)
+
           const dbOwner = asset.owner || asset.district_id || 'empty'
 
           log.info(`(asset-owner) id:(${asset.id})`)
@@ -92,11 +82,29 @@ const main = {
       .action(
         asSafeAction(async (assetId, assetType) => {
           const asset = await getAssetFromCLIArgs(assetId, assetType)
-          const contract = eth.getContract('LANDRegistry')
+          const contract = getContractByAssetType(assetType)
+
           const operator = await contract.updateOperator(asset.token_id)
 
           log.info(`(asset-update-operator) id:(${asset.id})`)
           log.info(`blockchain => ${operator}`)
+        })
+      )
+
+    program
+      .command('asset-data <assetId> <assetType>')
+      .description('Get the data for an asset')
+      .action(
+        asSafeAction(async (assetId, assetType) => {
+          const asset = await getAssetFromCLIArgs(assetId, assetType)
+          const contract = getContractByAssetType(assetType)
+
+          const data = await contract.tokenMetadata(asset.token_id)
+          const dbData = toDataLog(asset.data)
+
+          log.info(`(land-data) coords:(${x},${y})`)
+          log.info(`blockchain => ${data}`)
+          log.info(`db         => ${dbData}`)
         })
       )
 
@@ -112,24 +120,6 @@ const main = {
           const estateId = await contract.getLandEstateId(tokenId)
 
           log.info(`(land-estate) coords:(${x},${y}) => estateId: ${estateId}`)
-        })
-      )
-
-    program
-      .command('land-data <coord>')
-      .description('Get the land data for a (x,y) coordinate')
-      .action(
-        asSafeAction(async coord => {
-          const [x, y] = parseCLICoords(coord)
-          const contract = eth.getContract('LANDRegistry')
-          const data = await contract.landData(x, y)
-
-          const parcel = await Parcel.findOne({ x, y })
-          const dbData = toDataLog(parcel.data)
-
-          log.info(`(land-data) coords:(${x},${y})`)
-          log.info(`blockchain => ${data}`)
-          log.info(`db         => ${dbData}`)
         })
       )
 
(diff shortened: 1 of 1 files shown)


scripts/mktcli.js
```diff
@@ -48,23 +59,47 @@ const main = {
...
       .action(
-        asSafeAction(async coord => {
-          const [x, y] = parseCLICoords(coord)
-          const contract = eth.getContract('LANDRegistry')
-          const owner = await contract.ownerOfLand(x, y)
+        asSafeAction(async (assetId, assetType) => {
+          const asset = await getAssetFromCLIArgs(assetId, assetType)
+          let contract
+
+          switch (assetType) {
```
nachomazzara: What do you think of moving this to a helper function to retrieve the contract based on the `assetType`? we can use it in other places

scripts/mktcli.js
```diff
@@ -48,23 +59,47 @@ const main = {
...
         })
       )
 
+    program
+      .command('asset-update-operator <assetId> <assetType>')
+      .description('Get the update operator of an asset')
+      .action(
+        asSafeAction(async (assetId, assetType) => {
+          const asset = await getAssetFromCLIArgs(assetId, assetType)
+          const contract = eth.getContract('LANDRegistry')
```
nachomazzara: We should have this command for Estates too.

Related to [have a helper function to get the contract based on the assetType](https://github.com/decentraland/marketplace/pull/588/files#diff-14c145377d6748040b73ee576c8e1d57R69)

scripts/mktcli.js
```diff
@@ -130,36 +150,58 @@ const main = {
...
 
-          const pubDb = (await Publication.findByAssetId(id))[0]
+          // publication[1] === owner. See `toPublicationLog`
+          if (Contract.isEmptyAddress(publication[1].toString())) {
+            const landRegistry = eth.getContract('LANDRegistry')
+            const estateRegistry = eth.getContract('EstateRegistry')
+
+            contract = eth.getContract('Marketplace')
+
+            switch (assetType) {
```
nachomazzara: If we use [have a helper function to get the contract based on the assetType](https://github.com/decentraland/marketplace/pull/588/files#diff-14c145377d6748040b73ee576c8e1d57R69)

then we can do
```javascript
const contract = getContractByAssetType(assetType)
const publication = await contract.orderByAssetId(
                  contract.address,
                  asset.token_id
                )
```

scripts/mktcli.js
```diff
@@ -168,36 +210,38 @@ const main = {
...
-          const contract = eth.getContract('Marketplace')
+        asSafeAction(
+          async (assetId, assetType, contractName = 'Marketplace') => {
+            const asset = await getAssetFromCLIArgs(assetId, assetType)
+            const contract = eth.getContract(contractName)
 
-          await unlockAccountWithPrompt()
-          const txHash = await contract.cancelOrder(tokenId)
+            await unlockAccountWithPrompt()
+            const txHash = await contract.cancelOrder(asset.token_id)
```
nachomazzara: This will fail 😞 

Now the method is `function cancelOrder(address nftAddress, uint256 assetId)` we need the `address` of the asset

scripts/mktcli.js
```diff
@@ -168,36 +210,38 @@ const main = {
...
+        asSafeAction(async (assetId, assetType, options) => {
+          const asset = await getAssetFromCLIArgs(assetId, assetType)
           const events = BlockchainEvent.getEvents()
-          const [x, y] = parseCLICoords(coord)
-          const tokenId = await Parcel.encodeTokenId(x, y)
 
           const blockchainEvents = await BlockchainEvent.findByArgs(
             'assetId',
-            tokenId
+            asset.token_id
```
nachomazzara: Should we use `asset.id` instead? Doubt because of:  https://github.com/decentraland/marketplace/pull/588/files#diff-14c145377d6748040b73ee576c8e1d57R202

scripts/mktcli.js
```diff
@@ -244,29 +288,32 @@ const main = {
...
-        asSafeAction(async (coord, options) => {
-          const [x, y] = parseCLICoords(coord)
-          const tokenId = await Parcel.encodeTokenId(x, y)
-          const events = await BlockchainEvent.findByArgs('assetId', tokenId)
+        asSafeAction(async (assetId, assetType, options) => {
+          const asset = await getAssetFromCLIArgs(assetId, assetType)
+
+          const events = await BlockchainEvent.findByAnyArgs(
+            ['assetId', '_landId'],
+            asset.token_id
```
nachomazzara: same about `asset.id`

scripts/mktcli.js
NicoSantangelo: good catch!!

scripts/mktcli.js
NicoSantangelo: it's kind of confusing, but blockchain_events only deals with blockchain data it knows nothing about db `ids`

scripts/mktcli.js
NicoSantangelo: I like it, but where would you define it?

nachomazzara: As it is an isolated file I do it here

scripts/mktcli.js
```diff
@@ -101,35 +92,34 @@ const main = {
...
+        asSafeAction(async (assetId, assetType) => {
+          const asset = await getAssetFromCLIArgs(assetId, assetType)
+          const contract = getContractByAssetType(assetType)
 
-          const estateId = await contract.getLandEstateId(tokenId)
+          const data = await contract.tokenMetadata(asset.token_id)
+          const dbData = toDataLog(asset.data)
 
-          log.info(`(land-estate) coords:(${x},${y}) => estateId: ${estateId}`)
+          log.info(`(land-data) coords:(${x},${y})`)
```
nachomazzara: This could be also an Estate right? maybe we should change it to something more generic

scripts/mktcli.js
```diff
@@ -161,27 +151,14 @@ const main = {
...
-              }
-              case ASSET_TYPES.estate: {
-                publication = await contract.orderByAssetId(
-                  estateRegistry.address,
-                  asset.token_id
-                )
-                break
-              }
-            }
+            publication = await contract.orderByAssetId(
```
nachomazzara: What do you think of using the [legacy flag](https://github.com/decentraland/marketplace/pull/588/commits/192b97b43ed1421eac004210e3445f232f71dd65#diff-14c145377d6748040b73ee576c8e1d57R192) to get legacy active publication

NicoSantangelo: In this particular case, because it's a read, the command is checking for the legacy publication (`Contract.isEmptyAddress(publication[1].toString())`) and defaulting to the new one if it doesn't exist

We could have a problem if the asset has a publication in both contracts (unlikely but possible). But I think it might be better to just return everything instead of using a flag here.

Something like:

```
log.info(`(publication) id:(${asset.id})`)
log.info(`blockchain legacy => ${legacyPublication}`)
log.info(`blockchain        => ${publication}`)
log.info(`db                => ${publicationDb}`)
```

What do you think?

nachomazzara: As we are going to shut down legacy publications, it is going to be weird to have always the legacy log there. I still prefer to have the flag to decide when send the command in which marketplace I want to see the publication

NicoSantangelo: Hm but when we remove legacy publications this and the flag will be removed (both things).
Contrary to publication-cancel, I'm we're not making a destructive action which requires gas to work, the user needs information about an asset. If we add the flag here (a read) we're effectively making the user run it twice, once per each marketplace.
I want to know the publications associated with this asset, why should I know where that publication is?

I don't know if I'm really explaining this correctly

nachomazzara: Ahhh ok, I get the point. Agree with you now!

NicoSantangelo: Updated!

PR #589, opened 2018-10-18
fix: fat fingers

cazala:
Prevent selling parcels/estates for under 1000 MANA (this always happens by mistake)


NicoSantangelo:
🖐

PR #593, opened 2018-10-18
[DO NOT MERGE] perf: make inRange query >2x faster

NicoSantangelo:


migrations/1539867381972_publications-add-is-lastest-flag.js
```diff
@@ -0,0 +1,10 @@
+import { Publication } from '../src/Publication'
+
+exports.up = pgm => {
+  const tableName = Publication.tableName
+
+  pgm.addColumns(tableName, {
+    is_latest: { type: 'BOOLEAN', deafult: false }
```
abarmat: There is a typo here deafult => default

NicoSantangelo: 🤦‍♂️

migrations/1539867381972_publications-add-is-lastest-flag.js
```diff
@@ -0,0 +1,10 @@
+import { Publication } from '../src/Publication'
+
+exports.up = pgm => {
+  const tableName = Publication.tableName
+
+  pgm.addColumns(tableName, {
+    is_latest: { type: 'BOOLEAN', default: false }
+  })
+  pgm.addIndex(tableName, 'is_latest')
```
eordano: Missing:
```UPDATE from publications set is_latest=true
WHERE publications.id in (
  SELECT A.id from Publications A LEFT JOIN (
    SELECT id, asset_id from Publications GROUP BY asset_id SORT BY timestamp
  ) as B on A.id == B.id
)```

NicoSantangelo: Yes! @abarmat is testing this PR using a similar query to update current publications. I didn't add it to the PR to avoid re-writing it, and was waiting for him to do it. We were thinking of adding it as a script (`/scripts/update_is_latest.sql`) to avoid having to write the `down` part of the query, because it really depends on the state of the database at that particular moment.

eordano: oh, the down is just drop column and index :)

PR #597, opened 2018-10-19
feat: decentraland dapps for location, analytics and wallet

NicoSantangelo:


webapp/src/modules/wallet/actions.js
```diff
@@ -102,12 +74,12 @@ export function buyManaFailure(error) {
 
 // Update Balance
 
-export const UPDATE_BALANCE = 'Update Balance'
+export const UPDATE_MANA_BALANCE = 'Update Mana Balance'
```
nachomazzara: Should we need to run a migration for this change?

cazala: I don't think so, since the action creator `updateBalance` doesn't create a transaction action. 

The one that those so is `buyManaSuccess`.

NicoSantangelo: Exactly, I had the same concern but we're not using it

PR #625, opened 2018-10-30
feat: new sell phase. Auction PR: 1st

NicoSantangelo:
Closes #614 #608 

It adds:
- The auction page in the navbar *without* any features (yet)
- The contract authorization check mocking the implementation
- The feature flag `REACT_APP_FF_AUCTION` to toggle the auction features

PR #626, opened 2018-10-30
refactor: coordinates. Auction PR: 2nd

NicoSantangelo:
Pretty big refactor removing duplications over the coordinates management. It *doesn't* add any new features

It continues #625 , will rebase after that.


shared/coordinate.spec.js
```diff
@@ -0,0 +1,56 @@
+import { expect } from 'chai'
+import * as coordinates from './coordinates'
+
+describe('coordinates', function() {
+  describe('.validateCoordinate', function() {
+    it('should throw if the supplied coordinates are invalid', function() {
+      expect(() => coordinates.validateCoordinate('a,b')).to.throw(
+        'The coordinate "a,b" are not valid'
```
eordano: `s/are not/is not`

PR #627, opened 2018-10-31
feat: auction detail

### LINKED ISSUE #606: User will be presented with a map of Decentraland where it can select a number of unsold parcels
* User can select from 1 to N (adjacent) parcels for purchase.

* User will not be able to select parcels that currently have an owner. Note: To avoid any delay with having this data from the monitor, we can check directly with the contract from the frontend.

* User should see the selected LAND parcel for bidding.

* User should be able to deselect parcels not only from the map but also from the detail of parcels selected.

NicoSantangelo:
Adds the detail page for the auction, meaning what you see when you click the navbar auction. You can click unowned parcels and see how the UI reacts, all the data is mocked, so `Bid` doesn't work.

Closes #606 #610
Continues #626

### DIFF
--- a/shared/parcel.js
+++ b/shared/parcel.js
@@ -10,6 +10,16 @@ export function isParcel(asset) {
   )
 }
 
+export function getParcelSorter() {
+  return (parcel1, parcel2) => {
+    if (parcel1.x !== parcel2.x) {
+      return parcel1.x > parcel2.x ? -1 : 1
+    } else {
+      return parcel1.y > parcel2.y ? 1 : -1
+    }
+  }
+}
+
 export function toParcelObject(parcelArray, prevParcels = {}) {
   const parcelObject = {}
   for (const parcel of parcelArray) {
--- a/shared/parcel.js
+++ b/shared/parcel.js
@@ -10,16 +10,6 @@ export function isParcel(asset) {
   )
 }
 
-export function getParcelSorter() {
-  return (parcel1, parcel2) => {
-    if (parcel1.x !== parcel2.x) {
-      return parcel1.x > parcel2.x ? -1 : 1
-    } else {
-      return parcel1.y > parcel2.y ? 1 : -1
-    }
-  }
-}
-
 export function toParcelObject(parcelArray, prevParcels = {}) {
   const parcelObject = {}
   for (const parcel of parcelArray) {
--- a/src/Translation/locales/es.json
+++ b/src/Translation/locales/es.json
@@ -104,6 +104,7 @@
     "your_parcel": "Tu parcela"
   },
   "auction_page": {
+    "bid": "Oferta",
     "description": "Seleccione LAND para comprar",
     "maximum_parcels_message":
       "Si intenta comprar más de {max} parcelas en una sola transacción, fallará. ",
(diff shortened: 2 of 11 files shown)


shared/parcel.js
```diff
@@ -10,6 +10,16 @@ export function isParcel(asset) {
   )
 }
 
+export function getParcelSorter() {
```
nachomazzara: Just a thought: we could name it as `parcelSortAlgorithm` or just `sortParcel`

src/Translation/locales/es.json
```diff
@@ -104,6 +104,7 @@
     "your_parcel": "Tu parcela"
   },
   "auction_page": {
+    "bid": "Oferta",
```
nachomazzara: Maybe `Ofertar` is better

webapp/src/components/AuctionPage/AuctionPage.js
```diff
@@ -95,24 +115,57 @@ export default class AuctionPage extends React.PureComponent {
...
-                        parcel={parcel}
-                        withLink={false}
-                      />
-                    ) : null
-                  })}
+              </Grid.Column>
+              <Grid.Column mobile={16} computer={10}>
+                <div className="information-blocks">
+                  <div className="information-block">
+                    <p className="subtitle">GAS PRICE</p>
```
nachomazzara: Is it worth to have this on `translations`?

webapp/src/components/AuctionPage/AuctionPage.js
```diff
@@ -95,24 +115,57 @@ export default class AuctionPage extends React.PureComponent {
...
+                    <Header size="large">300wei</Header>
+                  </div>
+                  <div className="information-block">
+                    <p className="subtitle">PARCEL PRICE</p>
```
nachomazzara: Maybe `LAND PRICE` and also I would add it to the translations

webapp/src/components/AuctionPage/AuctionPage.js
```diff
@@ -95,24 +115,57 @@ export default class AuctionPage extends React.PureComponent {
...
+                    <Header size="large">{this.parcelPrice}</Header>
+                  </div>
+                  <div className="information-block">
+                    <p className="subtitle">PARCELS</p>
```
nachomazzara: Same about translations

webapp/src/components/AuctionPage/AuctionPage.js
```diff
@@ -95,24 +115,57 @@ export default class AuctionPage extends React.PureComponent {
...
+                    <Header size="large">
+                      {selected.length}/{MAX_PARCELS_PER_TX}
+                    </Header>
+                  </div>
+                  <div className="information-block">
+                    <p className="subtitle">TOTAL PRICE</p>
```
nachomazzara: Same about translations

shared/parcel.js
cazala: "sortParcel" would be the _return value_ of this function, since that's the actual function that does the sorting. I think it's fine returning it as a lambda tho, as it is doing now.

NicoSantangelo:
Updated!

NicoSantangelo:
Updated without sort

PR #632, opened 2018-11-01
feat: auction modal

NicoSantangelo:
Auction modal flow:

1) If you enter `/auction` for the first time and you don't have the contract authorized, you'll see a button redirecting you to Settings.
2) If you enter `/auction` for the first time and you do have the contract authorized, you'll have a button to start that dismisses the modal
3) If you enter `/auction` for the second time and you didn't interact with the modal you'll see 1 or 2 depending on the initial condition
4) If you enter `/auction` having accepted 2) you won't see anything

It also refactors the base modal a bit to keep it consistent, remove unused code and styles and working on all resolutions.

Closes #632 
Continues #626

### DIFF
--- a/src/Translation/locales/es.json
+++ b/src/Translation/locales/es.json
@@ -104,13 +104,13 @@
     "your_parcel": "Tu parcela"
   },
   "auction_modal": {
-    "authorize_contract": "Contrato de autorización",
-    "authorized_explanation": "Estás listo para comenzar! ",
+    "authorize_contract": "Autorizar contrato",
+    "authorized_explanation": "Estás listo para comenzar!",
     "blog_post": "Entrada en el blog",
     "first_time_explanation":
       "Para participar en esta nueva subasta, deberá autorizar el nuevo contrato {contract_link}. ",
-    "lets_go": "¡Vamonos!",
-    "title": "Decentraland Subasta"
+    "lets_go": "¡Adelante!",
+    "title": "Subasta de Decentraland"
   },
   "auction_page": {
     "bid": "Ofertar",
@@ -424,11 +424,11 @@
     "by_agreeing":
       "Al seleccionar \"Acepto\" a continuación, acepta nuestro {terms_of_service_link}.",
     "explanation":
-      "Este dApp interactúa con un intercambio descentralizado que se ejecuta en la cadena de bloques Ethereum.",
+      "Este dApp interactúa con un exchange descentralizado que se ejecuta en la cadena de bloques Ethereum.",
     "i_agree": "Estoy de acuerdo",
     "privacy_policy": "Política de privacidad",
     "terms_of_service": "Términos de servicio",
-    "title": "Bienvenido a Decentraland's LAND Marketplace"
+    "title": "Bienvenido al Mercado de LAND de Decentaland"
   },
   "transaction": {
     "buy": "Has comprado {parcel_link}",
--- a/src/Translation/locales/es.json
+++ b/src/Translation/locales/es.json
@@ -409,6 +409,18 @@
         "Puedes usar navegadores móviles como {coinbase_link} o {imtoken_link}."
     }
   },
+  "terms_modal": {
+    "also_agree":
+      "También acepta nuestro {privacy_policy_link}, que describe cómo procesamos su información.",
+    "by_agreeing":
+      "Al seleccionar \"Acepto\" a continuación, acepta nuestro {terms_of_service_link}.",
+    "explanation":
+      "Este dApp interactúa con un intercambio descentralizado que se ejecuta en la cadena de bloques Ethereum.",
+    "i_agree": "Estoy de acuerdo",
+    "privacy_policy": "Política de privacidad",
+    "terms_of_service": "Términos de servicio",
+    "title": "Bienvenido a Decentraland's LAND Marketplace"
+  },
   "transaction": {
     "buy": "Has comprado {parcel_link}",
     "buy_mana": "Has comprado {mana} MANA",
--- a/src/Translation/locales/es.json
+++ b/src/Translation/locales/es.json
@@ -103,6 +103,15 @@
     "your_estate": "Su propiedad",
     "your_parcel": "Tu parcela"
   },
+  "auction_modal": {
+    "authorize_contract": "Contrato de autorización",
+    "authorized_explanation": "Estás listo para comenzar! ",
+    "blog_post": "Entrada en el blog",
+    "first_time_explanation":
+      "Para participar en esta nueva subasta, deberá autorizar el nuevo contrato {contract_link}. ",
+    "lets_go": "¡Vamonos!",
+    "title": "Decentraland Subasta"
+  },
   "auction_page": {
     "bid": "Ofertar",
     "description": "Seleccione LAND para comprar",
--- a/webapp/src/components/AuctionPage/AuctionPage.js
+++ b/webapp/src/components/AuctionPage/AuctionPage.js
@@ -50,10 +50,14 @@ export default class AuctionPage extends React.PureComponent {
 
     if (
       !hasSeenAuctionModal() ||
-      (authorization && !isAuthorized(authorization))
+      this.isAuctionContractAuthorized(authorization)
     ) {
       onShowAuctionModal()
     }
+  }
+
+  isAuctionContractAuthorized(authorization) {
+    return authorization && !isAuthorized(authorization)
   }
 
   selectUnownedParcel = (asset, { x, y }) => {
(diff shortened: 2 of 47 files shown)


src/Translation/locales/es.json
```diff
@@ -103,6 +103,15 @@
     "your_estate": "Su propiedad",
     "your_parcel": "Tu parcela"
   },
+  "auction_modal": {
+    "authorize_contract": "Contrato de autorización",
+    "authorized_explanation": "Estás listo para comenzar! ",
```
nachomazzara: Oops., It is different from English translation and also there is an extra space at the end `"Estás listo para comenzar! "'

src/Translation/locales/es.json
```diff
@@ -103,6 +103,15 @@
     "your_estate": "Su propiedad",
     "your_parcel": "Tu parcela"
   },
+  "auction_modal": {
+    "authorize_contract": "Contrato de autorización",
```
nachomazzara: `Autorizar contrato`

src/Translation/locales/es.json
```diff
@@ -103,6 +103,15 @@
...
+    "authorized_explanation": "Estás listo para comenzar! ",
+    "blog_post": "Entrada en el blog",
+    "first_time_explanation":
+      "Para participar en esta nueva subasta, deberá autorizar el nuevo contrato {contract_link}. ",
+    "lets_go": "¡Vamonos!",
```
nachomazzara: where? hahahahahahahha

src/Translation/locales/es.json
```diff
@@ -103,6 +103,15 @@
...
+    "title": "Decentraland Subasta"
```
nachomazzara: `Subasta de Decentraland`

src/Translation/locales/es.json
```diff
@@ -405,6 +414,18 @@
         "Puedes usar navegadores móviles como {coinbase_link} o {imtoken_link}."
     }
   },
+  "terms_modal": {
+    "also_agree":
+      "También acepta nuestro {privacy_policy_link}, que describe cómo procesamos su información.",
+    "by_agreeing":
+      "Al seleccionar \"Acepto\" a continuación, acepta nuestro {terms_of_service_link}.",
+    "explanation":
+      "Este dApp interactúa con un intercambio descentralizado que se ejecuta en la cadena de bloques Ethereum.",
```
nachomazzara: `***Esta***...cadena de bloques ***de*** ....`

Maybe we should keep `intercambio` as `exchange`

src/Translation/locales/es.json
```diff
@@ -405,6 +414,18 @@
...
+    "i_agree": "Estoy de acuerdo",
+    "privacy_policy": "Política de privacidad",
+    "terms_of_service": "Términos de servicio",
+    "title": "Bienvenido a Decentraland's LAND Marketplace"
```
nachomazzara: `Bienvenido al Mercado de LAND de Decentaland` ?

webapp/src/components/AuctionPage/AuctionPage.js
```diff
@@ -25,6 +28,25 @@ export default class AuctionPage extends React.PureComponent {
...
+  componentWillReceiveProps(nextProps) {
+    this.showAuctionModal(nextProps)
+  }
+
+  showAuctionModal(props) {
+    const { authorization, onShowAuctionModal } = props
+
+    if (
+      !hasSeenAuctionModal() ||
+      (authorization && !isAuthorized(authorization))
```
nachomazzara: What do you think of moving`(authorization && !isAuthorized(authorization))` into a variable like

```javascript
const isAuctionContractAuthorized = authorization && !isAuthorized(authorization)
```

webapp/src/components/BuyEstatePage/BuyEstatePage.js
```diff
@@ -116,7 +113,11 @@ export default class BuyEstatePage extends React.PureComponent {
     const { isConnected, isLoading } = this.props
 
     if (isLoading) {
-      return <Loading />
+      return (
```
nachomazzara: Hmmm, Do you remember why you need to change it?

It would be great to just use `<Loading />` to show the spinner to avoid code duplication between components

webapp/src/components/ContractLink/ContractLink.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react'
```
nachomazzara: I like this component and I also have the same idea. But If I am not wrong, this component would be useful to have it at `decentraland-dapps`.

It is not part of the PR but we can create an issue to move it there

webapp/src/components/Modal/AuctionModal/AuctionModal.container.js
```diff
@@ -0,0 +1,21 @@
...
+import { navigateTo } from '@dapps/modules/location/actions'
+import { getAddress } from 'modules/wallet/selectors'
+import { getData as getAuthorizations } from 'modules/authorization/selectors'
+
+import AuctionModal from './AuctionModal'
+
+const mapState = state => {
+  const address = getAddress(state)
+  const authorization = getAuthorizations(state)[address]
+
```
nachomazzara: If we don't have a case where we care about authoriazations of not the current address. What do you think of moving `getAddress(state)` as part of the` getAuthorizations` selector?

We could avoid the need of doing:

```javascript
const address = getAddress(state)
const authorization = getAuthorizations(state)[address]
```

And do just 
```javascript
const authorization = getAuthorizations(state)
```

webapp/src/components/Modal/AuctionModal/AuctionModal.js
```diff
@@ -0,0 +1,113 @@
...
+  static propTypes = {
+    ...BaseModal.propTypes,
+    authorization: authorizationType,
+    onAuthorize: PropTypes.func.isRequired
+  }
+
+  handleOnAuthorize = () => {
+    const { onClose, onAuthorize } = this.props
+    onClose()
+    onAuthorize()
```
nachomazzara: Is it ok to close and authorize? I'm wondering if changing the order has an effect

src/Translation/locales/es.json
NicoSantangelo: HAHAH

webapp/src/components/ContractLink/ContractLink.js
NicoSantangelo: I like the idea, @cazala what do you think?

webapp/src/components/Modal/AuctionModal/AuctionModal.js
NicoSantangelo: it should be the same, the onClose only changes a boolean in the global redux state. I wanted to close it first because `onAuthorize` is a redirect. So you close and _then_ see something else.
In practice it's so fast that it's the same thing

webapp/src/components/Modal/AuctionModal/AuctionModal.container.js
NicoSantangelo: Hmm, we may need authorisations for another address, but I guess it's not that common. I'm not sure about coupling a method name like `getAuthorizations` with `getAddress` but I see your point.

I cast @cazala 🧙‍♂️. What do you think?

webapp/src/components/BuyEstatePage/BuyEstatePage.js
NicoSantangelo: My reasoning behind removing this is that this component looked like accidental duplication. It's only function was enabling `<Loader />` for `display: flex` parents.

That said, I can see that we're repeating:

`<div><Loader /></div>`

in quite a few places. I'm not sure if we should abstract this or not, if we do, I'd do it top level, but Loading sounds weird when we do use `Loader` in a couple of places. Once again @cazala  what do you think?

webapp/src/components/ContractLink/ContractLink.js
cazala: 👍 

There are a bunch of Etherscan related components that we have and could live as a component or family of components like BlockDate, EtherscanLink and this one

webapp/src/components/Modal/AuctionModal/AuctionModal.container.js
NicoSantangelo: Ok! I checked and we're already doing this in a couple of places so I'll go with that!

webapp/src/components/ContractLink/ContractLink.js
NicoSantangelo: https://github.com/decentraland/decentraland-dapps/issues/24

NicoSantangelo:
Updated!

nachomazzara:
We can merge it after rebase (L

PR #650, opened 2018-11-07
feat: show new events and fetch correctly

NicoSantangelo:
It's not trying to be 100% complete with the descriptions

### DIFF
--- a/scripts/mktcli.js
+++ b/scripts/mktcli.js
@@ -223,13 +223,13 @@ const main = {
         asSafeAction(async (assetId, assetType, options) => {
           const asset = await getAssetFromCLIArgs(assetId, assetType)
 
-          const blockchainEvents = await BlockchainEvent.findByArgs(
-            'assetId',
+          const blockchainEvents = await BlockchainEvent.findByAnyArgs(
+            ['assetId', 'landId', '_landId', '_estateId'],
             asset.token_id
           )
           blockchainEvents.reverse()
 
-          const eventLog = blockchainEvents.map(event => {
+          const eventLogPromises = blockchainEvents.map(async event => {
             const { name, block_number, log_index, tx_hash, args } = event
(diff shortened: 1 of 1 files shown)


scripts/mktcli.js
```diff
@@ -223,13 +223,13 @@ const main = {
         asSafeAction(async (assetId, assetType, options) => {
           const asset = await getAssetFromCLIArgs(assetId, assetType)
 
-          const blockchainEvents = await BlockchainEvent.findByArgs(
-            'assetId',
+          const blockchainEvents = await BlockchainEvent.findByAnyArgs(
+            ['assetId', 'landId', '_landId', '_estateId'],
```
nachomazzara: Also we have `_tokenId` for the ERC721 Transfer event

scripts/mktcli.js
```diff
@@ -290,7 +304,7 @@ const main = {
           const asset = await getAssetFromCLIArgs(assetId, assetType)
 
           const events = await BlockchainEvent.findByAnyArgs(
-            ['assetId', '_landId'],
+            ['assetId', 'landId', '_landId', '_estateId'],
```
nachomazzara: Maybe we could add also `_tokenId`

nachomazzara: Hmm What do you think if we create a const like

`const argsId = ['assetId', 'landId', '_landId', '_estateId', '_tokenId']`

IDK a good name: `argsId`, `wordToMatchId`, `matchId`, `argsNames`

NicoSantangelo: Agree with abstracting this!

NicoSantangelo:
Updated! let me know what you think of the constant name! (POSSIBLE_ASSET_EVENT_IDS)

nachomazzara:
giphy

PR #660, opened 2018-11-08
Feat: decentraland-eth breaking change

nachomazzara:
- Support overloaded methods

# Note to reviewer:

We should need to test all contract interactions


nachomazzara:
@NicoSantangelo fixed

PR #675, opened 2018-11-16
feat: use and show gas limit correctly. Fix tx

### LINKED ISSUE #674: Use gas price on transaction
Use gasPriceLimit to send the bid transaction

NicoSantangelo:
Closes #674


webapp/src/modules/auction/sagas.js
```diff
@@ -9,8 +10,12 @@ import {
   bidOnParcelsSuccess,
   bidOnParcelsFailure
 } from './actions'
+import { getParams } from './selectors'
+import { locations } from 'locations'
 import { splitCoodinatePairs } from 'shared/coordinates'
 
+const ONE_BILLION = 1000000000 // 1.000.000.000
```
cazala: 💰

PR #695, opened 2018-11-23
feat: finish auction page

### LINKED ISSUE #693: Finish Auction Page

nachomazzara:
## Note

It could be better to check #718 instead. It was branched from this branch and re-use the component created

Close #693 

### Tip for testing:

Change 
```javascript
export async function hasAuctionFinished() {
  const landAuction = eth.getContract('LANDAuction')
  const endTime = await landAuction.endTime()
  return endTime.toNumber() > 0 && Date.now() / 1000 >= endTime.toNumber()
}
```

To 

```javascript
export async function hasAuctionFinished() {
  const landAuction = eth.getContract('LANDAuction')
  const endTime = await landAuction.endTime()
  return endTime.toNumber() > 0 && Date.now() / 1000 >= endTime.toNumber() || true
}
```


webapp/src/modules/auction/utils.js
```diff
@@ -20,6 +21,8 @@ export const AUCTION_HELPERS = Object.freeze({
   SEEN_AUCTION_TOKEN_TOOLTIP: 'seenAuctionTokenTooltip'
 })
 
+export const AUCTION_DURATION_IN_DAYS = 15
```
cazala: is it 15 or 14?

nachomazzara: moved to the slack channel

nachomazzara:
Closed by #718

PR #699, opened 2018-11-23
fix: update auction copies

NicoSantangelo:


NicoSantangelo:
oh snap those conflicts

PR #700, opened 2018-11-23
feat: added authorization/confirmation step

### LINKED ISSUE #666: Authorize per token when the user clicks bid

### LINKED ISSUE #670: Add a modal to confirm a bid

### LINKED ISSUE #687: Remove `Auctioned LAND using MANA` from Settings

cazala:
Closes #666  
Closes #670 
Closes #687

I added the authorization/confirmation step. I also changed the settings page to allow users to unauthorize tokens authorized thru the auction page.


webapp/src/modules/authorization/sagas.js
```diff
@@ -77,8 +81,17 @@ function* handleAllowTokenRequest(action) {
     const contractToApprove = eth.getContract(contractName)
     const tokenContract = eth.getContract(tokenContractName)
 
+    // AUCTION
+    let options
+    const pathname = yield select(getPathname)
+    if (pathname === locations.auction()) {
+      const { gasPriceLimit } = yield select(getParams)
+      const gasPrice = gasPriceLimit * 1000000000
```
NicoSantangelo: Not sure about this one, but putting it out there.

I find the `* 1000000000` and `/ 1000000000` a bit confusing. `/auction/sagas` is doing it to (with `const ONE_BILLION = 1000000000`).

I don't want to add an utils function on `auction/utils` because `gasPrice` is not auction only. So right know I'd advise we keep an eye on this, specially if we decide to start hardcoding gas values on every tx

webapp/src/modules/wallet/utils.js
```diff
@@ -23,18 +23,83 @@ export function getWalletSagaOptions() {
...
     eth
   }
 }
 
-function getLandAuctionContract() {
+function getLandAuctionContracts() {
   const { LANDAuction } = contracts
 
+  // ZIL
+  const ZILToken = new contracts.ERC20Token(
```
NicoSantangelo: I think this is perfectly ok, but what do you think about having a`getNewERC20Token(name, address)` in this file and using it for this tokens as well as for the `RCNToken` ?

NicoSantangelo:
UI/UX comments:

While you have a pending transaction on the auction authorization modal:
- The cancel button doesn't really cancel anything and _might_ be confusing wording
- I think we should redirect to the /activity page if the user clicks on the loading, to mimic the settings form usage

webapp/src/components/Modal/BidConfirmationModal/BidConfirmationModal.container.js
```diff
@@ -0,0 +1,80 @@
...
+
+const mapDispatch = dispatch => ({
+  onSubmit: (parcels, beneficiary) =>
+    dispatch(bidOnParcelsRequest(parcels, beneficiary)),
+  onAuthorize: token =>
+    dispatch(
+      allowTokenRequest(
+        getTokenAmountToApprove(),
+        'LANDAuction',
+        token + 'Token'
```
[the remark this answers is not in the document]
cazala: done ✅

webapp/src/components/SettingsPage/SettingsForm/SettingsForm.js
```diff
@@ -53,35 +53,44 @@ export default class SettingsForm extends React.PureComponent {
...
-              contract_link: <ContractLink contractName={contractName} />,
-              symbol: token.getSymbolByContractName(tokenContractName)
-            }}
-          />
-        </div>
-      </Form.Field>
-    ))
+            <div className="title">
+              {t(`settings.for_buying_${contractName}`, {
+                token: tokenContractName.replace('Token', '') // remove the "Token" part
```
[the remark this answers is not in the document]
cazala: done ✅

webapp/src/modules/wallet/utils.js
cazala: done ✅

webapp/src/modules/authorization/sagas.js
cazala: 👍

cazala:
@NicoSantangelo 

> The warning seems off? I have enough balance.
comma (:

> Also the mana icon is not aligned with the text
Fixed

>  The delete transaction history prompt is off
Fixed

> Authorizing one token authorizes every token
I watched the vid but I didn't understand this one

> The color of the dropdown on the /marketplace page and languages in the footer are different from the one on auction:
Yes, this will be fixed with the transition to the ui kit

> The cancel button doesn't really cancel anything and might be confusing wording
I disabled it while the tx is pending

> I think we should redirect to the /activity page if the user clicks on the loading, to mimic the settings form usage

I don't know if we want to take the user away from this screen at this point. Maybe we can open an etherscan link on a separate window, but the original design doesn't intend that to be a button (specially the error state has a small "retry" button inside). We can discuss that for another PR.

abarmat:
❤️

PR #705, opened 2018-11-24
feat: glowing unowned parcels

cazala:
This was a proposal by mat to make unowned parcels "glow" while the user has nothing selected to hint which parcels are selectable.

[glow.mp4.zip](https://github.com/decentraland/marketplace/files/2612390/glow.mp4.zip)

### DIFF
--- a/shared/map/render/Map.js
+++ b/shared/map/render/Map.js
@@ -17,9 +17,11 @@ export class Map {
     estates,
     publications,
     selected,
-    wallet
+    wallet,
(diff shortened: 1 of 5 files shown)


NicoSantangelo:
I really like it. But I'm not sure it'll have the intended result. Do you think it makes sense to ask people not dev related for feedback ?

My only design comment is that I'd make the glow a bit less bright. I mean tone down the brightest value just a bit

🚀

PR #710, opened 2018-11-26
feat: defer update of the owner, update owners of selection, remove sold parcels from actual purchase

### LINKED ISSUE #689: Update selected parcels ownership
We should keep polling the ownership of the selected parcels and update them in case they get bought.

Sold parcels should still be shown in the list of parcels but they must not account for the total and must not be passed to the contract's bid method.

We shouldn't wait for the retrival of the owner on-chain when the user selects a parcel, it should be added to the list right away, and in case it turns out to be sold then the parcel should be changed to sold state

cazala:
This PR does a bunch of things:

+ defers the update of the owner of the selected parcel so it can be rendered as selected right away

+ refresh the owners of the selected parcels every 10 seconds (configurable)

+ removes sold parcels from the actual purchase (they are deducted from the total of parcels and the total price)

As you can see in this screen recording, the user can select parcels right away, and some of them change to "sold" status after selecting them. They are not accounted for the total price and total LAND amount. They still show up on the list to let the user know that those parcels have been sold, and the user can click on them to remove them from the list (either by clicking the X on the corner of the chip, or by clicking on the tile on the map). Once they are removed from the list, the user can't add them again by clicking on them.

689b

I also added small fixes for some issues:

+ Updated the colors of the SVGs in the sign in page

+ Fixed the total price shown in the confirmation modal

+ Updated the copy for "parcels available" as requested by ari.m

Closes #689

### DIFF
--- a/webapp/src/components/AuctionPage/AuctionPage.js
+++ b/webapp/src/components/AuctionPage/AuctionPage.js
@@ -70,6 +70,7 @@ export default class AuctionPage extends React.PureComponent {
 
     this.hasFetchedParams = false
     this.mounted = false
+    this.timoutId = null
 
     this.state = {
       showTokenTooltip: !hasSeenAuctionHelper(
@@ -95,6 +96,10 @@ export default class AuctionPage extends React.PureComponent {
 
   componentWillUnmount() {
     this.mounted = false
+    if (this.timoutId != null) {
+      clearTimeout(this.timeoutId)
+      this.timeoutId = null
+    }
   }
 
   componentWillReceiveProps(nextProps) {
@@ -199,7 +204,7 @@ export default class AuctionPage extends React.PureComponent {
 
   updateSelectionOwners = async () => {
     await Promise.all(this.getSelectedParcels().map(this.updateOwner))
-    setTimeout(() => {
+    this.timoutId = setTimeout(() => {
       if (!this.mounted) return
       this.updateSelectionOwners()
     }, REFRESH_OWNERS_INTERVAL)
--- a/webapp/src/components/AuctionPage/AuctionPage.js
+++ b/webapp/src/components/AuctionPage/AuctionPage.js
@@ -21,7 +21,8 @@ import {
   authorizationType,
   auctionParamsType,
   walletType,
-  parcelType
+  parcelType,
+  coordsType
 } from 'components/types'
 import {
   hasSeenAuctionHelper,
@@ -57,8 +58,10 @@ export default class AuctionPage extends React.PureComponent {
     onSetParcelOnChainOwner: PropTypes.func.isRequired,
     onFetchAvailableParcel: PropTypes.func.isRequired,
     onChangeAuctionCenterParcel: PropTypes.func.isRequired,
+    onChangeCoords: PropTypes.func.isRequired,
     token: PropTypes.oneOf(TOKEN_SYMBOLS),
-    rate: PropTypes.number
+    rate: PropTypes.number,
+    selectedCoordinatesById: PropTypes.objectOf(coordsType).isRequired
   }
 
   constructor(props) {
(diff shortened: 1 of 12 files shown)


NicoSantangelo:
Two things:

- I find a bit weird that I can de-select a SOLD parcel by clicking the map again, because it's no longer "selected". Still, super non critical
deselect

- If I select a _soon to be SOLD_ parcel and click PURCHASE, the parcel gets updated but the confirmation doesn't:

image
purchase

NicoSantangelo:
Also, this might be intentional but when you click `PROCEED` the modal closes before you `accept/reject` the transaction. So, you can keep moving around but if you do accept the TX you'll get redirected:

go-away

NicoSantangelo:
If I pick a soon to be sold parcel as the 20th parcel in my cart, the Warn sign doesn't go away once it gets updated:

image

webapp/src/components/AuctionPage/AuctionPage.js
```diff
@@ -168,6 +184,22 @@ export default class AuctionPage extends React.PureComponent {
...
+    return this.getParcelOwnerOnChain(parcel).then(ownerOnChain => {
+      if (!Contract.isEmptyAddress(ownerOnChain)) {
+        this.props.onSetParcelOnChainOwner(parcel.id, ownerOnChain)
+      }
+    })
+  }
+
+  updateSelectionOwners = async () => {
+    await Promise.all(this.getSelectedParcels().map(this.updateOwner))
+    setTimeout(() => {
```
NicoSantangelo: _Question_: maybe we sould store this timeoutId and clear it on unmount, even if it doesn't do anything with the `if (!this.mounted) return`. I'm thinking it might be cumbersome if the user enters and leaves the /auction page repeatedly

webapp/src/components/AuctionPage/AuctionPage.js
```diff
@@ -243,7 +280,10 @@ export default class AuctionPage extends React.PureComponent {
       )
     }
 
-    const selectedParcels = this.getSelectedParcels()
+    const selectedParcels = this.getSelectedParcels(true)
```
NicoSantangelo: unused true?
```suggestion
    const selectedParcels = this.getSelectedParcels()
```

cazala:
@NicoSantangelo 

> I find a bit weird that I can de-select a SOLD parcel by clicking the map again, because it's no longer "selected". Still, super non critical

I added it as an alternative of clicking on the 'X' because it seemed to me that the SOLD status looked a bit like disabled, so in case it wasn't intuitive enough they could also deselect by clicking on it (it happend to me actually that I though I could do it and it didn't work, so I added it). I don't think this is a must have but I don't think it hurts to have it, we can discuss this with the rest of the team, if we don't want it then I can remove the feat.

> If I select a soon to be SOLD parcel and click PURCHASE, the parcel gets updated but the confirmation doesn't:

Fixed by https://github.com/decentraland/marketplace/pull/710/commits/0b4e9bc61c6bf09f824891de267d57f831eb7abc

> Also, this might be intentional but when you click PROCEED the modal closes before you accept/reject the transaction. So, you can keep moving around but if you do accept the TX you'll get redirected:

Fixed by https://github.com/decentraland/marketplace/pull/710/commits/435d11c2b7290538ea6a0fef08c9fb25982961f6

> If I pick a soon to be sold parcel as the 20th parcel in my cart, the Warn sign doesn't go away once it gets updated:

Fixed by https://github.com/decentraland/marketplace/pull/710/commits/ab9af8ed863162b2ed6c9e7021950a5a587a9390

> When I click on a parcel the owner doesn't get updated until the 10-second timer goes off.

Fixed by https://github.com/decentraland/marketplace/pull/710/commits/0b4e9bc61c6bf09f824891de267d57f831eb7abc

webapp/src/components/AuctionPage/AuctionPage.js
cazala: ✅ done https://github.com/decentraland/marketplace/pull/710/commits/23d917a67410008d925d134506142de74ec61a12

webapp/src/components/AuctionPage/AuctionPage.js
cazala: ✅ done https://github.com/decentraland/marketplace/pull/710/commits/231badd70903ce30b18c8a2acc6ba6b37355648f

PR #715, opened 2018-11-27
feat: See more parcels + Collapsable + Chip

### LINKED ISSUE #701: Add a see-more parcels on the Estate detail after a certain amount

NicoSantangelo:
Closes #701


webapp/src/components/Chip/Chip.js
```diff
@@ -2,62 +2,59 @@ import React from 'react'
...
   getClassName() {
-    const { size, status, onClick, onDelete } = this.props
+    const { className, size, status, onClick, onDelete } = this.props
     const statusClass = status
       ? `has-status status-${status.toLowerCase()}`
       : ''
     const onClickClass = onClick ? 'clickeable' : ''
     const onDeleteClass = onDelete ? 'deleteable' : ''
-    return `ParcelCoord ${size} ${statusClass} ${onClickClass} ${onDeleteClass}`
+    return `Chip ${className} ${size} ${statusClass} ${onClickClass} ${onDeleteClass}`
```
nachomazzara: Just to be consistent with the rest of the variables. What do you think of adding

```javascript
const sizeClass = size
```

webapp/src/components/Collapsable/Collapsable.js
```diff
@@ -0,0 +1,83 @@
...
+      isCollapsable: false
+    }
+  }
+
+  toggleSeeMore = () => {
+    this.setState({ isCollapsed: !this.state.isCollapsed })
+  }
+
+  componentWillMount() {
+    window.addEventListener('resize', this.debouncedSetCollapsableUsingHeight)
```
nachomazzara: I couldn't see the `see more` until I resize the window. I think that we should call `setCollapsableUsingHeight` once we mount this component and keep the listener just in case the user resize the window.

Also, I think that we need to add a `padding` to the `auction-panel`

screen shot 2018-11-27 at 15 18 30

webapp/src/components/Chip/Chip.js
```diff
@@ -2,62 +2,59 @@ import React from 'react'
...
     status: ''
   }
 
   handleOnClick = () => {
-    const { onClick, parcel } = this.props
-    if (onClick) {
-      onClick(parcel)
-    }
+    const { onClick } = this.props
+    if (onClick) onClick()
```
nachomazzara: I think that we are trying to always use brackets even though it has only one line

webapp/src/components/EditEstatePage/EstateSelect/EstateSelect.css
```diff
@@ -7,55 +7,34 @@
...
   text-align: left;
   font-size: 14px;
   margin-top: 15px;
   color: var(--text);
 }
+
+@media (max-width: 475px) {
+  .EstateSelect .estate-selection .dissolve-button {
+    padding-bottom: 6px;
+    float: none;
```
nachomazzara: Question: Did you want to use `clear: both` here?

webapp/src/components/EditEstatePage/EstateSelect/EstateSelect.js
```diff
@@ -201,52 +212,42 @@ export default class EstateSelect extends React.PureComponent {
...
+                  />
+                </Grid.Column>
+              )}
               {canEdit && (
-                <Grid.Column
-                  className="parcel-actions-container"
-                  computer={8}
-                  tablet={16}
-                >
+                <Grid.Column width={14}>
```
nachomazzara: Should it be `16`?

16

screen shot 2018-11-27 at 15 40 03

14

screen shot 2018-11-27 at 15 40 08

webapp/src/components/EditEstatePage/EstateSelect/EstateSelect.js
```diff
@@ -201,52 +212,42 @@ export default class EstateSelect extends React.PureComponent {
                       max: MAX_PARCELS_PER_TX
                     })}
                   />
-                  <p className="warning-parcels-limit" />
                 </Grid.Column>
               </Grid.Row>
             ) : null}
 
             <Grid.Row>
-              <Grid.Column width={isCreation ? 16 : 8}>
```
nachomazzara: Good catch! that was obsolete!

webapp/src/components/EditEstatePage/EstateSelect/EstateSelectActions/EstateSelectActions.js
```diff
@@ -1,9 +1,9 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import { Button } from 'semantic-ui-react'
+import { t } from '@dapps/modules/translation/utils'
```
nachomazzara: ❤️

webapp/src/components/ParcelCoords/ParcelCoord/ParcelCoord.css
```diff
@@ -0,0 +1,2 @@
+.ParcelCoord {
+}
```
nachomazzara: Hmmm, Could we remove this file?

webapp/src/components/ParcelCoords/ParcelCoords.js
```diff
@@ -0,0 +1,38 @@
...
+    isCollapsable: PropTypes.bool
+  }
+
+  static defaultProps = {
+    isCollapsable: true
+  }
+
+  render() {
+    const { children, parcels, isCollapsable, ...parcelCoordProps } = this.props
+    const maxHeight = isCollapsable ? MAX_PARCEL_COORD_HEIGHT : Infinity
```
nachomazzara: As `Collapsable` receive the `maxHeight` can we just send it as `null` if `isCollapsable = false` and remove `maxHeight: PropTypes.number.isRequired,` from Collapsable component. 

It will display `maxHeight: 100%` if not valid height is passed. Receive it as `Infinity` makes me go again to check what happen with a big height is passed

webapp/src/components/EditEstatePage/EstateSelect/EstateSelect.css
NicoSantangelo: Nope! I wanted to remove the `float: right` from the button.
If I clear the float I'd look like this:

image

but I want this:

image

webapp/src/components/ParcelCoords/ParcelCoords.js
NicoSantangelo: You're right, it's not clear what the `Infinity` is doing here. So instead of having the `Collapsable` receive a `null` I just check on the father component if I have to have a `Collapsable`.

So instead of `isCollapsable ? MAX_PARCEL_COORD_HEIGHT : Infinity` we'd do
`return isCollapsable ? <Collapsable>div</Collapsable> : div`
which also saves us components and logic

webapp/src/components/Collapsable/Collapsable.js
NicoSantangelo: Added the padding!
The see more on the land auction wasn't supposed to be there 😅 fixed

PR #718, opened 2018-11-28
feat: auction countdown

### LINKED ISSUE #698: Homepage banner to showcase new Auction

nachomazzara:
This PR is based on #695 following the same approach and using the component in a reusable way.

Closes #698


webapp/src/components/AuctionFinishedPage/AuctionFinishedPage.js
```diff
@@ -0,0 +1,82 @@
...
+export default class AuctionFinishedPage extends React.PureComponent {
+  static propTypes = {
+    onGoToMarketplace: PropTypes.func.isRequired,
+    isConnected: PropTypes.bool.isRequired,
+    isConnecting: PropTypes.bool.isRequired,
+    params: auctionParamsType
+  }
+
+  getAuctionRealDuration = () => {
+    const { startTime, endTime } = this.props.params
```
nachomazzara: get `startTime` from `auction/utils`

NicoSantangelo: - Add a `// The transaction that ends the auction might take a while to finish so we cap the max duration`comment
- Move it to utils

webapp/src/components/AuctionRoute/AuctionRoute.js
```diff
@@ -0,0 +1,70 @@
...
+  async componentWillReceiveProps(nextProps) {
+    if (nextProps.isConnected) {
+      this.fetchAuctionParams()
+      const auctionFinished = await hasAuctionFinished()
+      const auctionHasStarted = hasAuctionStarted()
+      this.setState({
+        auctionFinished,
+        auctionHasStarted
+      })
+    }
```
NicoSantangelo: maybe we can extract
 
```
      this.fetchAuctionParams()
      const auctionFinished = await hasAuctionFinished()
      const auctionHasStarted = hasAuctionStarted()
      this.setState({
        auctionFinished,
        auctionHasStarted
      })
    }
```

to it's own function?

webapp/src/components/AuctionSplash/AuctionSplash.js
```diff
@@ -0,0 +1,38 @@
...
+
+  render() {
+    const { onSubmit } = this.props
+
+    return (
+      <AuctionStaticPage>
+        <div className="AuctionSpash">
+          <AuctionCountdown>
+            <p>{t('auction_splash.cta_title')}</p>
+            <Form onSubmit={onSubmit} className="cta-wrapper">
```
NicoSantangelo: I think need to add preventDefault from `lib/utils` here?
```javascript
import { preventDefault } from 'lib/utils'
```

```suggestion
            <Form onSubmit={preventDefault(onSubmit)} className="cta-wrapper">
```

webapp/src/components/AuctionStaticPage/AuctionStaticPage.css
```diff
@@ -0,0 +1,21 @@
+.Page {
+  position: relative;
```
NicoSantangelo: Whoops! this will add `position: relative` on _all_ pages !

webapp/src/components/Countdown/Countdown.css
```diff
@@ -0,0 +1,34 @@
...
+.Countdown-col-element.dots {
+  margin: 0;
+}
+
+.Countdown-col-element p, .banner .Countdown-col-element p{
+  font-size: 2em;
+  margin-bottom: 0;
+}
+
+Countdown-col-element span {
```
NicoSantangelo:
```suggestion
.Countdown-col-element span {
```

?

webapp/src/components/Countdown/Countdown.css
```diff
@@ -0,0 +1,34 @@
...
+  text-transform: capitalize
+}
+
+@media(min-width: 768px) {
+  .Countdown-col-element{
+    margin: 0 20px;
+  }
+
+  .Countdown-col-element p{
```
NicoSantangelo: Can I ask for a prettier format to this file?

for example:
```suggestion
  .Countdown-col-element p {
```

webapp/src/components/Countdown/Countdown.js
```diff
@@ -0,0 +1,139 @@
...
+      min: 0,
+      sec: 0
+    }
+  }
+
+  componentDidMount() {
+    // update every second
+    this.interval = setInterval(() => {
+      const date = this.calculateCountdown(this.props.date)
+      date ? this.setState(date) : this.stop()
```
NicoSantangelo: This is correct, but I read date and got confused. Was about to point out that you where setting a date in `setState()`. This happens not only because of the name `date` itself, but because the prop `date` is a number

Maybe `const timeLeft = this.calculateCountdown(this.props.date)` (borrowing from calculateCountdown) or `const newTimeLeftState = this.calculateCountdown(this.props.date)`

webapp/src/components/Countdown/Countdown.js
```diff
@@ -0,0 +1,139 @@
...
+
+    return timeLeft
+  }
+
+  stop() {
+    clearInterval(this.interval)
+  }
+
+  addLeadingZeros(value) {
+    value = String(value)
```
NicoSantangelo: _Question_: Why not `.toString()`? there are a couple of differences with `String()` but I think this is weird here:

`String(undefined) => "undefined"`

webapp/src/components/Countdown/Countdown.js
```diff
@@ -0,0 +1,139 @@
...
+  render() {
+    const countDown = this.state
+
+    return (
+      <div className="Countdown">
+        <span className="Countdown-col">
+          <span className="Countdown-col-element">
+            <p>{this.addLeadingZeros(countDown.days)}</p>
+            <span>
+              {countDown.days === 1 ? t('global.day') : t('global.days')}
```
NicoSantangelo: _Not actionable_
I wouldn't change this, but if you have to add more text here, like for example `Ends in 1 day` or `Ends in 2 days` remember that you can use our translate ""engine"" here:

`"countdown.days": "Ends in {days} {days, plural, one {day} other {days}}`
`t('countdown.days', { days: countDown.days })`

I know, it's horrible

webapp/src/components/Countdown/Countdown.js
```diff
@@ -0,0 +1,139 @@
...
+            <p>:</p>
+          </span>
+        </span>
+
+        <span className="Countdown-col">
+          <span className="Countdown-col-element">
+            <p>{this.addLeadingZeros(countDown.sec)}</p>
+            <span>{t('global.seconds')}</span>
+          </span>
+        </span>
```
NicoSantangelo: This HTML looks super weird to me, too many `span`s for no reason? It's not that's bad or anything but the building block for `xxx-col` seems like it should be a div?

cc: @cazala

webapp/src/components/Countdown/Countdown.js
```diff
@@ -0,0 +1,139 @@
...
+  addLeadingZeros(value) {
+    value = String(value)
+    while (value.length < 2) {
+      value = '0' + value
+    }
+    return value
+  }
+
+  render() {
+    const countDown = this.state
```
NicoSantangelo: Why not:
```suggestion
    const { days, hours, secs, ...} = this.state
```
?

webapp/src/components/HomePage/HomePage.js
```diff
@@ -19,6 +22,11 @@ export default class HomePage extends React.PureComponent {
     this.props.onFetchPublications()
   }
 
+  onLearnMore = () => {
+    const { onNavigate } = this.props
+    onNavigate(locations.auction())
```
NicoSantangelo: I know `onNavigate` was already on the container but I think it was legacy (it wasn't being used). I think it's a bit better to delegate the `locations.auction` to the container and delete it:

```suggestion
    this.props.onNavigateToAuction()
```

**BUT**, maybe the `<Button primary={true} onClick={this.onLearnMore}>` can just be a link?

```
<Link to={locations.auction()}>
  <Button primary={true}>
```

So in conclusion, using a link and removing the method from the container

webapp/src/modules/auction/reducer.js
```diff
@@ -81,18 +85,18 @@ export function auctionReducer(state = INITIAL_STATE, action) {
...
-            landsLimitPerBid: params.landsLimitPerBid,
-            currentPrice: params.currentPrice
-          }
+          params: Object.keys(params).reduce(
+            (acc, key) => ({
+              ...acc,
+              [key]: params[key]
+            }),
+            {}
+          )
```
NicoSantangelo: This is 👍 , but I wanted to explain. The reasoning behind not using a reduce here (in mind) was not setting incorrect params by accident.
Maybe we can use 

```javascript
utils.pick(params, Object.keys(INITIAL_STATE.data.params))
```

what do you think?

webapp/src/modules/auction/sagas.js
```diff
@@ -129,9 +137,12 @@ function* handleConnectWalletSuccess(action) {
   // keep refreshing params and rate while the user is on /auction
   while (connected) {
     const pathname = yield select(getPathname)
-    if (pathname === locations.auction()) {
+    const isActive = yield call(() => isAuctionActive())
```
NicoSantangelo: _Consistency_
```suggestion
    const isActive = yield call(isAuctionActive)
```

webapp/src/components/Countdown/Countdown.js
nachomazzara: Just something to have in mind. I copied and paste it.

I could revisit it. It was done to speed up the development. My bad

webapp/src/components/Countdown/Countdown.js
nachomazzara: Idk if I get it.

We only show `Day` or `Days` not a full phrase like `ends in...` or `starts in...`

I think the phrase could be in a title like

Ends in
{{Countdown here}}

Or

Starts in
{{countdown here}}

webapp/src/Routes.js
```diff
@@ -158,7 +158,7 @@ export default class Routes extends React.Component {
 
         {/*Auction*/}
         {isFeatureEnabled('AUCTION') && (
-          <Route exact path={locations.auction()} component={AuctionPage} />
+          <Route exact path={locations.auction()} component={AuctionRoute} />
```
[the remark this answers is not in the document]
nachomazzara: What do you thing of `AuctionRouterPage` or `AuctionPageRouter` just to have less impact on naming change

PR #739, opened 2018-12-02
Update ja.json

EijiGard:
last update for the auction!

### DIFF
--- a/src/Translation/locales/ja.json
+++ b/src/Translation/locales/ja.json
@@ -114,17 +114,17 @@
   "auction_modal": {
     "authorization": "承認",
     "authorization_description":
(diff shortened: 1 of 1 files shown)


src/Translation/locales/ja.json
```diff
@@ -266,7 +266,7 @@
     "expired_at": "{time}に締切済み",
     "expires_in": "締切：{time}",
     "expires_in_more_than_years": "締切まであと{years}年以上",
-    "for": "for",
+    "for": " ",
```
nachomazzara: Just to double check. Is this ok?

EijiGard: Yes. I see It's only used in a sentence that "you're about buying X,Y for 000MANA". It can't be translated "for" as itself in Japanese. It's understandable even without it:)

PR #744, opened 2018-12-04
feat: accept all terms from auction modal

### LINKED ISSUE #743: Accept both terms and conditions on the auction page
If the user didn't accept the general terms and conditions she should be able to accept both (changing the copy) via the auction modal

cazala:
Closes #743 

Users who haven't accepted Decentraland's Terms of Service and Privacy Policy will see this checkbox instead of the one that only links to the Auction terms:

screen shot 2018-12-05 at 3 01 12 pm

And by checking and confirming that modal they won't see the other ToS modal on the rest of the marketplace.

### DIFF
--- a/webapp/src/components/Modal/AuctionModal/AuctionModal.js
+++ b/webapp/src/components/Modal/AuctionModal/AuctionModal.js
@@ -72,6 +72,7 @@ export default class AuctionModal extends React.PureComponent {
         href="https://decentraland.org/terms-auction"
         rel="noopener noreferrer"
         target="_blank"
+        onClick={this.hanleClickExternalLink}
       >
         {t('auction_modal.auction_terms_link')}
       </a>
@@ -84,6 +85,7 @@ export default class AuctionModal extends React.PureComponent {
         href="https://decentraland.org/terms"
         rel="noopener noreferrer"
         target="_blank"
+        onClick={this.hanleClickExternalLink}
       >
         {t('auction_modal.decentraland_terms_link')}
       </a>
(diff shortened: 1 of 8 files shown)


webapp/src/components/Modal/AuctionModal/AuctionModal.js
```diff
@@ -88,8 +124,6 @@ export default class AuctionModal extends React.PureComponent {
   }
 
   renderWelcome() {
-    const { hasAgreedToTerms } = this.state
```
NicoSantangelo: _Silly comment_ We're using `this.state.` a few times below, maybe destructuring works here?

webapp/src/components/Modal/AuctionModal/AuctionModal.js
```diff
@@ -3,6 +3,10 @@ import PropTypes from 'prop-types'
 import { Checkbox, Button } from 'semantic-ui-react'
 import { t, T } from '@dapps/modules/translation/utils'
 
+import {
+  hasAgreedToTerms as hasAgreedToOtherTerms,
+  agreeToTerms as agreeToOtherTerms
```
NicoSantangelo: _Silly comment_ Maybe instead of `Other` we could use something like `Global` or `Main`? I.e `hasAgreedToMainTerms`

webapp/src/components/Modal/AuctionModal/AuctionModal.js
cazala: this is because `hasAgreedToMainTerms` can't be deconstructed cos there's a function with that name

PR #748, opened 2018-12-05
feat: added banner for started and finished auction

### LINKED ISSUE #741: Check what happens when the countdown reaches 0. Consider refreshing the page or adding a CTA

cazala:
Closes #741

This PR makes the countdown refresh the page when it reaches zero. Also I added a state for when the auction is already started and another one for when it has finished. This is for the homepage (so the banner doesn't disappear).


NicoSantangelo:
One more thing, if I go to the auction page from the Homepage, going back using the browser (history) keeps all the modals open except for the countdown

PR #777, opened 2018-12-10
feat: check balance before purchase

cazala:
Show balance on confirmation modal and show warning + disable proceed button if it's not enough

screen shot 2018-12-10 at 4 05 03 pm


webapp/src/components/Modal/BidConfirmationModal/BidConfirmationModal.js
```diff
@@ -24,6 +32,60 @@ export default class BidConfirmationModal extends React.PureComponent {
...
+      const contract = eth.getContract(contractName)
+      const balance = await contract.balanceOf(address)
+
+      this.setState({
+        balance:
+          balance && typeof balance.toNumber === 'function'
+            ? balance.toNumber()
+            : balance,
+        isLoading: false
+      })
```
NicoSantangelo: should we worry about this failing too? (returning 0) I don't know how ofter etherscan fails of if it has a rate limit, asking just to be safe because we're disabling the button

webapp/src/components/Modal/BidConfirmationModal/BidConfirmationModal.js
```diff
@@ -1,15 +1,23 @@
...
 import { parcelType } from 'components/types'
-import { TOKEN_SYMBOLS } from 'modules/auction/utils'
+import { TOKEN_SYMBOLS, TOKEN_ADDRESSES } from 'modules/auction/utils'
+import { token } from 'lib/token'
 import BaseModal from '../BaseModal'
 
 import './BidConfirmationModal.css'
 
+function formatAmount(amount = 0, digits = 2) {
+  return parseFloat(amount.toFixed(digits), 10).toLocaleString()
```
NicoSantangelo: I think we have a `parseFloatWithDecimal` on auction/utils

webapp/src/components/Modal/BidConfirmationModal/BidConfirmationModal.js
```diff
@@ -24,6 +32,60 @@ export default class BidConfirmationModal extends React.PureComponent {
...
+    }
+  }
+
+  fetchBalance = async (symbol, address) => {
+    this.setState({ balance: null, isLoading: true })
+
+    try {
+      const contractAddress = TOKEN_ADDRESSES[symbol]
+      const response = await axios.get(
+        `https://api.etherscan.io/api?module=account&action=tokenbalance&contractaddress=${contractAddress}&address=${address}&tag=latest&apikey=${env.get(
```
NicoSantangelo: shouldn't we check the network here?
Also maybe we can move it to it's own method or object like EtherscanAPI

webapp/src/components/Modal/BidConfirmationModal/BidConfirmationModal.js
```diff
@@ -24,6 +32,60 @@ export default class BidConfirmationModal extends React.PureComponent {
...
+          'REACT_APP_ETHERSCAN_API_KEY'
+        )}`
+      )
+      const balance =
+        Number(response.data.result) / (symbol === 'ZIL' ? 10 ** 12 : 10 ** 18)
```
nachomazzara: We are using `parseInt(value, 10)` in every places. Maybe we could use it here too

webapp/src/components/Modal/BidConfirmationModal/BidConfirmationModal.js
```diff
@@ -24,6 +32,60 @@ export default class BidConfirmationModal extends React.PureComponent {
...
+      }
+      this.setState({
+        balance,
+        isLoading: false
+      })
+    } catch (error) {
+      window.Rollbar.info('Fetching balance via Infura')
+      const contractName = token.getContractNameBySymbol(symbol)
+      const contract = eth.getContract(contractName)
+      const balance = await contract.balanceOf(address)
```
nachomazzara: Should we need to convert from _wei_ like `balane / (symbol === 'ZIL' ? 10 ** 12 : 10 ** 18)`

We could have `let balance` and format it after the catch

webapp/src/components/Modal/BidConfirmationModal/BidConfirmationModal.js
```diff
@@ -48,14 +111,35 @@ export default class BidConfirmationModal extends React.PureComponent {
...
 
+        {!isLoading && balance != null ? (
+          <React.Fragment>
+            <div className="your-balance">
+              {t('auction_modal.your_balance', {
+                balance: formatAmount(balance),
+                token
+              })}
+            </div>
+            {balance < price ? (
```
nachomazzara: Hmmm I don't know if we are already doing it, but I think we should change the price with the margin (fee applied)

webapp/src/components/Modal/BidConfirmationModal/BidConfirmationModal.js
```diff
@@ -48,14 +111,35 @@ export default class BidConfirmationModal extends React.PureComponent {
...
+          </React.Fragment>
+        ) : null}
+
         <div className="actions">
-          <Button primary onClick={this.handleSubmit}>
-            {t('auction_modal.submit')}
+          <Button
+            primary
+            onClick={this.handleSubmit}
+            disabled={price > balance || isLoading}
```
nachomazzara: Maybe we could move `balance < price` to a `const hasEnoughBalance = balance >= price`

webapp/src/components/Modal/BidConfirmationModal/BidConfirmationModal.js
cazala: i didn't get any errors fetching from etherscan, i left there just in case, but we can remove it all together if you think it makes it messier.

webapp/src/components/Modal/BidConfirmationModal/BidConfirmationModal.js
cazala: 👍

webapp/src/components/Modal/BidConfirmationModal/BidConfirmationModal.js
cazala: done 👍

webapp/src/components/Modal/BidConfirmationModal/BidConfirmationModal.js
cazala: done 👍

webapp/src/components/Modal/BidConfirmationModal/BidConfirmationModal.js
cazala: done 👍

webapp/src/components/Modal/BidConfirmationModal/BidConfirmationModal.js
cazala: Wouldn't it be weird that we show one price before hitting purchase and a different one on the confirmation modal? I we do so, I think we should show the price with margin as the big price next to the purchase button

webapp/src/components/Modal/BidConfirmationModal/BidConfirmationModal.js
cazala: done 👍

webapp/src/components/Modal/BidConfirmationModal/BidConfirmationModal.js
nachomazzara: I will do it cause if we let the user proceed with less than the amount he should have, it will fail.

Is it ok to do this in this PR? @cazala @NicoSantangelo

PR #789, opened 2018-12-13
feat: buy auction lands script

NicoSantangelo:
`./scripts/buyAuctionLands.js buy --help`
```
  Usage: buy [options]

  Options:

    --parcels [parcels]                    List of parcels to deploy. Required
    --manaTokenAddress [manaTokenAddress]  MANAToken contract address. Required
    --password [password]                  Password for the account
    --retryFailedTxs                       If this flag is present, the script will try to retry failed transactions
    -h, --help                             output usage information
```

`./scripts/buyAuctionLands.js buy --parcels ./scripts/parcels.json --manaTokenAddress 0x72fd6C7C1397040A66F33C2ecC83A0F71Ee46D5c`

where parcels.json:
```
[
  {x, y}
]
```


scripts/buyAuctionLands.js
```diff
@@ -0,0 +1,198 @@
...
+    }
+    parcels.push(parcel)
+  }
+
+  return parcels
+}
+
+async function bidOnParcels(parcels, tokenAddress, shouldRetry = false) {
+  const landAuctionContract = eth.getContract('LANDAuction')
+  const account = eth.getAddress()
```
abarmat: Is it possible to use a different account?

scripts/buyAuctionLands.js
```diff
@@ -0,0 +1,198 @@
...
+
+  const gasPrice = (await landAuctionContract.gasPriceLimit()).toNumber()
```
abarmat: This is nice

scripts/buyAuctionLands.js
```diff
@@ -0,0 +1,198 @@
...
+  const landLimit = (await landAuctionContract.landsLimitPerBid()).toNumber()
+  log.info(
+    `Using ${gasPrice} as gasPrice and ${landLimit} as land limit per bid`
+  )
+
+  const txs = []
+  const parcelsToRetry = []
+
+  log.info(`Buying ${parcels.length} parcels with token ${tokenAddress}`)
+  await asyncBatch({
```
abarmat: Can we wait some time for N txs to be confirmed/rejected before sending new ones? This would work as other scripts.

scripts/buyAuctionLands.js
NicoSantangelo: I could add a `--account` and use it to call `web3.unlockAccount` directly and on the txs `from`. I thought it wasn't going to be necessary

scripts/buyAuctionLands.js
NicoSantangelo: Yes! I actually forgot. The script is waiting for each transaction to finish but after sending all transactions

NicoSantangelo:
Updated with `--account` and `--txDelay` !

scripts/buyAuctionLands.js
```diff
@@ -0,0 +1,236 @@
...
+        parcelsToRetry.push({ x, y: ys[index] })
+      }
+    }
+  }
+
+  log.info(`Bought ${parcels.length - parcelsToRetry.length} parcels`)
+
+  if (parcelsToRetry.length > 0 && shouldRetry) {
+    log.info(`Retrying on ${parcelsToRetry.length} parcels`)
+    return bidOnParcels(...args)
```
cazala: isn't this retrying with the same parcels as before, instead of using only the parcels in  `parcelsToRetry`?

scripts/buyAuctionLands.js
```diff
@@ -0,0 +1,236 @@
...
+  }
+}
+
+async function getConfirmedTransaction(hash, isRetrying = false) {
```
cazala: From the data we collected from Segment, 10 seconds might not be enough for this in Mainnet :/ 

I would suggest using something like

```javascript

async function getConfirmedTransaction(hash, retries = 0) {
...

if (retries < 3) {

...

await utils.sleep(10000)
return getConfirmedTransaction(hash, ++retries)

```

abarmat: I'm not sure, but if this mechanism is the same as the one used in the assignNewParcel script, it works!

PR #814, opened 2019-01-17
fix: update owner when transfer an estate

nachomazzara:
Update tiles owner when transferring an Estate

### DIFF
--- a/src/Tile/Tile.model.js
+++ b/src/Tile/Tile.model.js
@@ -48,7 +48,7 @@ export class Tile extends Model {
 
   static async upsertEstate(estate) {
     return asyncBatch({
-      elements: estate.data.parcels,
+      elements: estate ? estate.data.parcels : [], // Some Estates appear as undefined
       callback: async parcelsBatch => {
         const promises = []
 
--- a/src/Tile/Tile.model.js
+++ b/src/Tile/Tile.model.js
@@ -34,21 +34,26 @@ export class Tile extends Model {
   ]
 
   static async upsertAsset(assetId, assetType) {
+    // getNew throws if the assetType is invalid
     const asset = await Asset.getNew(assetType).findById(assetId)
 
+    if (!asset) {
+      // Some Estates appear as undefined
+      return
+    }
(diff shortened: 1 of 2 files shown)


src/Tile/Tile.model.js
```diff
@@ -48,7 +48,7 @@ export class Tile extends Model {
 
   static async upsertEstate(estate) {
     return asyncBatch({
-      elements: estate.data.parcels,
+      elements: estate ? estate.data.parcels : [], // Some Estates appear as undefined
```
NicoSantangelo: Do we know when this happens? because it looks like it should be a check in the caller. For example adding a 

`if (!asset) return`

on `upsertAsset`

nachomazzara: I would add it to `upsertAsset`, but also remove the `throw` related to the `assetType`

```javascript
default:
          throw new Error(`The asset type ${assetType} is invalid`)
```

It is weird to not throw if the asset is `undefined` or `null`, but the type is not `parcel` or `estate`:
```javascript
 static async upsertAsset(assetId, assetType) {
    const asset = await Asset.getNew(assetType).findById(assetId)
    if (asset) {
      // Some Estates appear as undefined
      switch (assetType) {
        case ASSET_TYPES.parcel:
          return this.upsertParcel(asset)
        case ASSET_TYPES.estate:
          return this.upsertEstate(asset)
        default:
          throw new Error(`The asset type ${assetType} is invalid`)
      }
    }
  }
```

I suggest, to omit if the asset is not valid: asset itself and type


```javascript
 static async upsertAsset(assetId, assetType) {
    const asset = await Asset.getNew(assetType).findById(assetId)
    if (asset) {
      // Some Estates appear as undefined
      switch (assetType) {
        case ASSET_TYPES.parcel:
          return this.upsertParcel(asset)
        case ASSET_TYPES.estate:
          return this.upsertEstate(asset)
        default:
         break
      }
    }
  }
```

abarmat: I'll try to reproduce on Ropsten

PR #820, opened 2019-01-23
feat: bid ui

nachomazzara:
webapp .env:

- `REACT_APP_ERC721_BID_CONTRACT_ADDRESS=`
- `REACT_APP_FF_BIDS`


webapp/src/components/AssetTransactionHistory/AssetTransactionHistory.js
```diff
@@ -39,21 +61,21 @@ export default class AssetTransactionHistory extends React.PureComponent {
         <Link to={locations.profilePageDefault(address)}>
           <AddressBlock
             address={address}
-            scale={4}
+            scale={3.4}
             hasTooltip={false}
             hasLink={false}
           />&nbsp;
-          <span className="short-address">{shortenAddress(address)}</span>
+          <span className="short-address">{shortenOwner(address)}</span>
```
nicosantangelo: Why are we doing this? is it a CSS problem? I think it can be avoided changing the column sizes:

image


that fixes it, keeping shortenAddress

webapp/src/components/AssetTransactionHistory/AssetTransactionHistory.js
```diff
@@ -7,25 +7,47 @@ import { locations } from 'locations'
...
+      .map(listing => ({
+        block_number: listing.block_number,
+        price: listing.price,
+        block_time_created_at: listing.block_time_created_at,
+        block_time_updated_at: listing.block_time_updated_at,
+        id: listing.tx_hash || listing.id,
+        from: listing.owner || listing.seller,
+        to: listing.buyer || listing.bidder,
+        type: listing.tx_hash ? LISTING_TYPES.PUBLICATION : LISTING_TYPES.BID
+      }))
```
nicosantangelo: The only way of telling if a listing is a bid or a publication is via the tx_hash prop? It scares me a bit because tomorrow we can add a tx_hash prop to a bid and this would break.

Don't you think it's a good idea to at least make this two separate methods to ease maintainability?

Something _like_ this:

`.map(listing => listing.tx_hash ? this.buildPublication(listing) : this.buildBid(listing))`

webapp/src/components/Bid/Bid.js
```diff
@@ -0,0 +1,68 @@
...
+  static propTypes = {
+    isOwner: PropTypes.bool.isRequired,
+    bid: bidType.isRequired
+  }
+
+  render() {
+    const { bid, isOwner, onConfirm, onUpdate } = this.props
+
+    return (
+      <Grid stackable className={'Bid'}>
```
nicosantangelo:
```suggestion
      <Grid stackable className="Bid">
```

webapp/src/components/Bid/Bid.js
```diff
@@ -0,0 +1,68 @@
...
+            <div className="address-wrapper" title={bid.bidder}>
+              <Link to={locations.profilePageDefault(bid.bidder)}>
+                <AddressBlock
+                  address={bid.bidder}
+                  scale={3.4}
+                  hasTooltip={false}
+                  hasLink={false}
+                />&nbsp;
+                <span className={`short-address ${!isOwner ? 'you' : ''}`}>
+                  {isOwner ? shortenAddress(bid.bidder) : t('global.you')}
```
nicosantangelo: I think it's a bit weird that the same condition is used in two different ways here, `isOwner ?` and `!isOwner ?`. Maybe we could use `isOwner ?` both times?

webapp/src/components/BidAssetPage/BidAssetForm.js
```diff
@@ -0,0 +1,184 @@
...
+      expiresAt: bid
+        ? dateFnsFormat(parseInt(bid.expires_at, 10), INPUT_FORMAT)
+        : this.formatFutureDate(DEFAULT_DAY_INTERVAL),
+      formErrors: []
+    }
+  }
+
+  handlePriceChange = e => {
+    // Dots and commas are not allowed
+    const price = e.currentTarget.value.replace(/\.|,/g, '')
```
nicosantangelo: This is repeated in three files already, maybe we could abstract it?

webapp/src/components/BidAssetPage/BidAssetForm.js
```diff
@@ -0,0 +1,184 @@
...
+    if (formErrors.length === 0) {
+      onBid({
+        asset_id: asset.id,
+        expires_at: this.toUTCTimestamp(expiresAt),
+        price: parseInt(price)
+      })
+    } else {
+      this.setState({ formErrors })
+    }
+  }
```
nicosantangelo: Same goes for these error checks, they're practically the same as PublishAssetForm. This abstraction is not **that** easy, so it can be added later but we should add an issue!

webapp/src/components/BidAssetPage/BidAssetPage.container.js
```diff
@@ -0,0 +1,100 @@
...
+    assetType: ownProps.assetType,
+    isTxIdle: isBidIdle(state),
+    isConnected: isConnected(state),
+    isLoading: isConnecting(state) || isLoading(state)
+  }
+}
+
+const mapDispatch = (dispatch, ownProps) => {
+  let onCancel
+  let onBid
```
nicosantangelo: const getOnBid = assetType => bid => dispatch(
          openModal('FatfingerModal', {
            onSubmit: () =>
              dispatch(bidRequest({ ...bid, asset_type: assetType })),

            priceToConfirm: bid.price
          })

webapp/src/modules/bid/sagas.js
```diff
@@ -0,0 +1,164 @@
...
+    const estates = yield select(getEstates)
+    const asset = yield call(() => buildAsset(asset_id, asset_type, estates))
+    const owner = yield select(getAddress)
+
+    let contract
+    if (asset_type === ASSET_TYPES.parcel) {
+      contract = eth.getContract('LANDRegistry')
+    } else {
+      contract = eth.getContract('EstateRegistry')
+    }
```
nicosantangelo: I think we can use `getContractByAssetType` here

webapp/src/modules/bid/sagas.js
```diff
@@ -0,0 +1,164 @@
...
+    yield put(push(locations.activity()))
+  } catch (error) {
+    yield put(acceptBidFailure(error.message))
+  }
+}
+
+function* handleFetchBidById({ bidId }) {
+  try {
+    const bids = yield call(() => api.fetchBidById(bidId))
+    yield put(fetchBidByIdSuccess(bids[0]))
```
nicosantangelo: Why does `fetchBidById` return an array? shouldn't it return a single bid?

webapp/src/modules/estates/reducer.js
```diff
@@ -147,6 +151,21 @@ export function estatesReducer(state = INITIAL_STATE, action) {
...
+      const estates = assets.filter(asset => isEstate(asset))
+      return {
+        ...state,
+        loading: loadingReducer(state.loading, action),
+        error: null,
+        data: {
+          ...state.data,
+          ...estates.reduce((acc, estate) => {
+            return { ...acc, [estate.id]: normalizeEstate(estate) }
+          }, state.data)
```
nicosantangelo: Maybe we can use a lambda here?

shared/listing.js
```diff
@@ -7,10 +7,23 @@ export const LISTING_STATUS = Object.freeze({
 
 export const LISTING_ASSET_TYPES = Object.freeze({
   parcel: 'parcel',
-  estate: 'estate',
-  unknow: 'unknow'
+  estate: 'estate'
+  // unknow: 'unknow'
```
nicosantangelo: Maybe we can remove this line?

shared/parcel.js
```diff
@@ -96,3 +97,14 @@ export function getParcelsNotIncluded(newParcels, allParcels) {
...
 }
+
+/**
+ * Check if parcel is listable or not
+ * @param parcel
+ * @return boolean - whether is listable or not
+ */
+export function isListable(parcel) {
+  return (
+    !isDistrict(parcel) && !isRoad(parcel) && !isPlaza(parcel) && parcel.owner
```
nicosantangelo: I think only `isDistrict` is necessary here, as it only check for the existence of a `district_id` prop
```suggestion
    !isDistrict(parcel) && parcel.owner
```

src/Listing/Bid/Bid.model.js
```diff
@@ -26,6 +28,21 @@ export class Bid extends Model {
...
+    )
+  }
+
+  static async findByAddressAndStatus(address, status) {
+    return this.db.query(
+      SQL`SELECT * from ${raw(this.tableName)}
+        WHERE ${BidQueries.bidderOrSeller(address)} 
+          AND status = ${status}`
+    )
+  }
```
nicosantangelo: Silly request, but to be consistent could we write the queries like this?:

```js
SQL`SELECT *
  FROM ${raw(this.tableName)}
  WHERE token_address = ${tokenAddress}
    AND token_id = ${tokenId}
    AND status = ANY(${statuses})`
```

src/Listing/Bid/Bid.model.js
```diff
@@ -81,12 +106,77 @@ export class Bid extends Model {
...
+        SET seller = ${seller}, block_time_updated_at = ${blockTime}, block_number = ${blockNumber}
+        WHERE block_time_created_at <= ${blockTime}
+          AND token_address = ${tokenAddress} 
+          AND asset_id = ${assetId}
+          AND status = ANY(${[
+            LISTING_STATUS.open,
+            LISTING_STATUS.fingerprintChanged
+          ]})`
+    )
+  }
```
nicosantangelo: I think we should try to refactor the arguments of:
- invalidateBids
- updateAssetByFingerprintChange
- updateAssetOwner

The queries themselves seem ok, but those methods take 4+ arguments each and it's hard to distinguish what each of them do. Just adding an arg to the end of a function is usually a bad idea if you already have 3.

Maybe we can group a few of those into what they represent? Also we could try to keep the API of db.update where we have `(changes, conditions)`.

Plus I see a few repeated conditions like

```js
WHERE block_time_created_at <= ${blockTime}
  AND token_address = ${tokenAddress}
  AND token_id = ${tokenId}
```

but abstracting those might be going too far. I'll leave that decision to you

src/Listing/Bid/Bid.model.js
```diff
@@ -81,12 +106,77 @@ export class Bid extends Model {
...
+
+  static async findBiddedAssetsByStatus(filters) {
```
nicosantangelo: - The past tense of bid is bid
- This method is intended to have a `status` arg?

src/Listing/Bid/Bid.model.js
```diff
@@ -81,12 +106,77 @@ export class Bid extends Model {
...
+    const { status, bidder /*, seller  */ } = filters
+    const Assets = new Listing.getListableAssets()
+
+    // @nacho TODO: temp code. move to estate.queries or asset
```
nicosantangelo: I think the only section of this query we should extract is the `EXISTS`. Can we do it on this PR?

src/Listing/Bid/Bid.model.js
```diff
@@ -81,12 +106,77 @@ export class Bid extends Model {
...
+    // prettier-ignore
+    const estateHasParcels = () =>
```
nicosantangelo: Is this a function because you plan to extract it? if so it's ok, but it could be a const otherwise

src/Listing/Bid/Bid.model.js
```diff
@@ -81,12 +106,77 @@ export class Bid extends Model {
...
+
+  static async findBiddedAssetsByStatus(filters) {
+    const { status, bidder /*, seller  */ } = filters
+    const Assets = new Listing.getListableAssets()
+
+    // @nacho TODO: temp code. move to estate.queries or asset
+    // prettier-ignore
+    const estateHasParcels = () => 
+     SQL`(bid.asset_type = ${LISTING_ASSET_TYPES.estate}
+      AND EXISTS(SELECT 1 FROM ${raw('parcels')} AS p WHERE p.estate_id = bid.asset_id)
```
nicosantangelo:
```suggestion
      AND EXISTS(SELECT 1 FROM ${raw(Parcel.tableName)} AS p WHERE p.estate_id = bid.asset_id)
```

src/Listing/Bid/Bid.model.js
```diff
@@ -81,12 +106,77 @@ export class Bid extends Model {
...
+    // Keep only the model that had a bid defined from the Assets list
+    assets = assets.map(asset => {
+      for (const Model of Assets) {
+        if (asset[Model.tableName] != null) {
+          Object.assign(asset, asset[Model.tableName])
+        }
+        delete asset[Model.tableName]
+      }
+      return asset
+    })
```
nicosantangelo: This method looks dangerously similar to `Marketplace`'s `filterAll` do you think it'd be possible to extract some common logic? like the asset `map`ing.

We might need a new concept/abstraction

src/Listing/Bid/Bid.router.js
```diff
@@ -82,17 +101,69 @@ export class BidRouter {
...
+    try {
+      const bidder = server.extractFromReq(req, 'bidder')
+      bids = bids.filter(bid => bid.bidder === bidder)
+    } catch (errror) {
+      // Do nothing
+    }
+
+    try {
+      const sanitize = server.extractFromReq(req, 'sanitize')
+      return sanitize === 'false' ? bids : sanitizeBids(bids)
```
nicosantangelo: You can maybe use `ReqQueryParams`'s `getBoolean` here

src/Listing/Bid/Bid.router.js
```diff
@@ -82,17 +101,69 @@ export class BidRouter {
...
+      const status = server.extractFromReq(req, 'status')
+      bids = await Bid.findByAddressAndStatus(address, status)
+    } catch (error) {
+      bids = await Bid.findByAddress(address)
+    }
+
+    return sanitizeBids(bids)
+  }
+
+  async getAssetBidded(req) {
```
nicosantangelo: The past tense of bid is bid

src/Listing/Bid/Bid.router.js
```diff
@@ -82,17 +101,69 @@ export class BidRouter {
...
+    const reqQueryParams = new ReqQueryParams(req)
+
+    const opt = {}
+    if (reqQueryParams.has('bidder')) {
+      opt['bidder'] = reqQueryParams.get('bidder')
+    }
+
+    if (reqQueryParams.has('status')) {
+      opt['status'] = reqQueryParams.get('status')
+    }
```
nicosantangelo: Creating `opt` is not necessary if status is not present on the query string, maybe we can move it below line 159

src/Listing/Bid/Bid.router.js
```diff
@@ -82,17 +101,69 @@ export class BidRouter {
...
+
+    let assets
+
+    try {
+      const status = server.extractFromReq(req, 'status')
+
+      assets = await Bid.findBiddedAssetsByStatus(opt, status)
```
nicosantangelo: `findBiddedAssetsByStatus` doesn't have a second argument

src/Listing/Bid/Bid.router.js
```diff
@@ -82,17 +101,69 @@ export class BidRouter {
...
+    } catch (error) {
+      console.log(error.message)
+      // assets = await Bid.findBiddedAssets(opt)
+    }
+
+    return sanitizeAssets(assets)
```
nicosantangelo: **General file question**:

I don't get why are we having sanitize as an argument for these endpoints. Granted I'm reading this top to bottom so maybe reading the front-end part will illuminate me, but can you explain?
Why not returning everything always? We never needed this before

src/ReqQueryParams/MarketplaceReqQueryParams.js
```diff
@@ -1,4 +1,5 @@
 import { ReqQueryParams } from './ReqQueryParams'
+
```
nicosantangelo: Intentional newline?

src/Translation/locales/en.json
```diff
@@ -32,7 +55,10 @@
   },
   "asset_cancel": {
     "about_to_cancel": "You are about to cancel the sale of {name}.",
+    "about_to_cancel_bid": "You are about to cancel the bid for {name}.",
+    "cancel_bid": "Cancel Bid",
     "cancel_sale": "Cancel Sale",
+    "not_bidded": "You don't have an open bid for the {asset_name} {asset}.",
```
nicosantangelo: The past tense of bid is bid

src/sanitize/sanitizeAssets.js
```diff
@@ -13,6 +14,9 @@ export function sanitizeAssets(assets) {
       ? sanitizePublication(publication)
       : undefined
 
+    //@nacho TODO: revisit- this
+    newAsset.bid = bids ? sanitizeBid(bids) : undefined
```
nicosantangelo: I don't get this line. `bids` becomes `bid` ? Shouldn't it be
```suggestion
    newAsset.bids = bids ? sanitizeBids(bids) : []
```

`sanitizeBid` only omits an `object`'s properties

webapp/src/Routes.js
```diff
@@ -89,6 +90,33 @@ export default class Routes extends React.Component {
           path={this.addLegacySupport(locations.parcelMapDetail())}
           component={AtlasPage}
         />
+        {isFeatureEnabled('BIDS') && (
```
nicosantangelo: This is OK but maybe we can add `{/*Bids*/}` and group this routes with the estates ones as well?
I think it'll look a bit more organised because parcels and estates already have a ton of routes, what do you think?

webapp/src/components/AcceptBidAssetPage/AcceptBidAssetPage.js
```diff
@@ -0,0 +1,46 @@
...
+export default class AcceptBidAssetPage extends React.PureComponent {
+  static propTypes = {
+    id: PropTypes.string.isRequired,
+    bid: bidType,
+    assetType: PropTypes.string.isRequired,
+    isConnected: PropTypes.bool,
+    isLoading: PropTypes.bool.isRequired
+  }
+
+  bidFetched = false
```
nicosantangelo:
```suggestion
  isBidFetched = false
```

I know it's not super good english, but to keep consistent

webapp/src/components/AcceptBidAssetPage/AcceptBidAssetPage.js
```diff
@@ -0,0 +1,46 @@
...
+
+  componentWillReceiveProps(nextProps) {
+    const { isConnected, isLoading, onFetchBidByAsset, bid } = nextProps
+
+    if (!isConnected || isLoading || bid || this.bidFetched) {
+      return
+    }
+    this.bidFetched = true
+
+    onFetchBidByAsset()
```
nicosantangelo: **Question**

Why not do this on `componentWillMount`?

webapp/src/components/AcceptBidAssetPage/AcceptBidAssetPage.js
```diff
@@ -0,0 +1,46 @@
...
+    switch (assetType) {
+      case ASSET_TYPES.parcel:
+        return <AcceptBidParcelPage {...this.props} />
+      case ASSET_TYPES.estate:
+        return <AcceptBidEstatePage {...this.props} />
+      default:
+        return null
+    }
+  }
+}
```
nicosantangelo: I noticed that the render (not renderPage) and handleConfirm methods of each `AcceptBid...Page` are the same. Maybe they belong on this file?

webapp/src/components/AcceptBidAssetPage/AcceptBidParcelPage/AcceptBidParcelPage.js
```diff
@@ -0,0 +1,88 @@
...
+import { t, T } from '@dapps/modules/translation/utils'
+
+import { splitCoordinate } from 'shared/coordinates'
+import ParcelModal from 'components/ParcelModal'
+import ParcelDetailLink from 'components/ParcelDetailLink'
+import Parcel from 'components/Parcel'
+import Mana from 'components/Mana'
+import { walletType, bidType } from 'components/types'
+
+// import './BuyParcelPage.css'
```
nicosantangelo: Whoops

webapp/src/components/AcceptBidAssetPage/AcceptBidParcelPage/AcceptBidParcelPage.js
```diff
@@ -0,0 +1,88 @@
...
+                })}
+                subtitle={
+                  <T
+                    id="asset_accept_bid.about_to_accept_bid"
+                    values={{
+                      name: <ParcelDetailLink parcel={parcel} />,
+                      price: (
+                        <React.Fragment>
+                          &nbsp;{t('global.for')}&nbsp;&nbsp;
+                          <span>
```
nicosantangelo: Same comments as `AcceptBidEstatePage`

webapp/src/components/Asset/Asset.js
```diff
@@ -33,8 +33,8 @@ export default class Asset extends React.PureComponent {
   }
 
   componentWillMount() {
-    const { value, isLoading, onFetchAsset } = this.props
-    if (value || isLoading) {
+    const { isLoading, onFetchAsset } = this.props
```
nicosantangelo: Could you remind me why are we changing this? it's a fairly delicate piece of code :P

cc: @cazala

webapp/src/components/AssetTransactionHistory/AssetTransactionHistory.container.js
```diff
@@ -0,0 +1,19 @@
+import { connect } from 'react-redux'
+
+import { fetchAssetListingHistory } from 'modules/asset//actions'
+import { getAcceptedBidsByAssetFactory } from 'modules/bid/selectors'
+import AssetTransactionHistory from './AssetTransactionHistory'
+
+const mapState = (state, ownProps) => {
+  const getbids = getAcceptedBidsByAssetFactory(ownProps.asset)
```
nicosantangelo: Why the lowercase here?

_I left this comment on modules/bid/selectors but careful this is not working as a memoized selector_

webapp/src/components/AssetTransactionHistory/AssetTransactionHistory.js
nicosantangelo: The only way of telling if a listing is a bid or a publication is via the tx_hash prop? It scares me a bit because tomorrow we can add a tx_hash prop to a bid and this would break.

Don't you think it's a good idea to at least make this two separate methods to ease maintainability?

Something _like_ this:

`.map(listing => listing.tx_hash ? this.buildPublication(listing) : this.buildBid(listing))`

webapp/src/components/AssetTransactionHistory/AssetTransactionHistory.js
nicosantangelo: Why are we doing this? is it a CSS problem? I think it can be avoided changing the column sizes:

image


that fixes it, keeping shortenAddress

webapp/src/components/Bid/Bid.css
```diff
@@ -0,0 +1,52 @@
...
+  width: 100%;
+}
+
+.Bid .row h3 {
+  text-transform: uppercase;
+  color: var(--gray);
+  font-weight: 300;
+}
+
+.Bid .row .address-wrapper a{
```
nicosantangelo:
```suggestion
.Bid .row .address-wrapper a {
```

webapp/src/components/Bid/Bid.js
nicosantangelo:
```suggestion
      <Grid stackable className="Bid">
```

webapp/src/components/Bid/Bid.js
nicosantangelo: I think it's a bit weird that the same condition is used in two different ways here, `isOwner ?` and `!isOwner ?`. Maybe we could use `isOwner ?` both times?

webapp/src/components/BidAssetPage/BidAssetForm.js
nicosantangelo: This is repeated in three files already, maybe we could abstract it?

webapp/src/components/BidAssetPage/BidAssetForm.js
nicosantangelo: Same goes for these error checks, they're practically the same as PublishAssetForm. This abstraction is not **that** easy, so it can be added later but we should add an issue!

webapp/src/components/BidAssetPage/BidAssetPage.container.js
nicosantangelo: **IF** you want to abstract onBid, you could do:

```js
  const getOnBid = assetType => bid =>
    dispatch(
      openModal('FatfingerModal', {
        assetType,
        priceToConfirm: bid.price,
        onSubmit: () => dispatch(bidRequest({ ...bid, asset_type: assetType }))
      })
    )
```

`onBid = getOnBid(ASSET_TYPES.parcel)`

only if you feel it'll help readability

webapp/src/components/BidAssetPage/BidEstatePage/BidEstatePage.js
```diff
@@ -0,0 +1,124 @@
...
+
+    if (isLoading) {
+      return this.renderLoading()
+    }
+
+    return (
+      <Estate id={id} ownerNotAllowed>
+        {estate => {
+          const isERC721BidAllowed =
+            authorization && authorization.allowances.ERC721Bid.MANAToken
```
nicosantangelo: I don't think this is the best way to go about this. It's technically correct, but adding a variable definition and return inside a component feels wrong somehow.

I think it's much better to just abstract this in a function/Component

cc: @cazala

webapp/src/components/BidAssetPage/BidParcelPage/BidParcelPage.js
```diff
@@ -0,0 +1,130 @@
...
+
+    if (isLoading) {
+      return this.renderLoading()
+    }
+
+    return (
+      <Parcel x={x} y={y} ownerNotAllowed>
+        {parcel => {
+          const isERC721BidAllowed =
+            authorization && authorization.allowances.ERC721Bid.MANAToken
```
nicosantangelo: Same as [here](https://github.com/decentraland/marketplace/pull/820/files#diff-e94a471b4f2bd7089cd0c2b7cb593726R56)

webapp/src/components/AcceptBidAssetPage/AcceptBidAssetPage.container.js
```diff
@@ -0,0 +1,72 @@
...
+  getBidById,
+  isCancelIdle,
+  isLoading as isBidLoading
+} from 'modules/bid/selectors'
+import AcceptBidAssetPage from './AcceptBidAssetPage'
+
+const mapState = (state, ownProps) => {
+  let assetId
+  const { bidId, id } = getMatchParams(ownProps)
+  assetId = id
```
nicosantangelo: You can just let here
```suggestion
  let assetId = id
```

webapp/src/components/AcceptBidAssetPage/AcceptBidAssetPage.container.js
```diff
@@ -0,0 +1,72 @@
...
+
+  switch (ownProps.assetType) {
+    case ASSET_TYPES.parcel: {
+      const { x, y } = getMatchParamsCoordinates(ownProps)
+      assetId = buildCoordinate(x, y)
+      break
+    }
```
nicosantangelo: I think we should be consistent between map dispatch and props, and I like the approach below.

It think it's clearer to have a case for estates here like:

```js
const matchParams = getMatchParams(ownProps)

// (...)
case ASSET_TYPES.estate: {
  assetId = matchParams.id
  break
}
```

instead of using id as default. You can use `matchParams.bidId` below

webapp/src/components/AcceptBidAssetPage/AcceptBidAssetPage.container.js
```diff
@@ -0,0 +1,72 @@
...
+    isLoading: isBidLoading(state),
+    isConnected: isConnected(state),
+    id: assetId
+  }
+}
+
+const mapDispatch = (dispatch, ownProps) => {
+  let onCancel
+
+  const { bidId } = getMatchParams(ownProps)
```
nicosantangelo: I'd invert this variable declaration to avoid newlines

```
  const { bidId } = getMatchParams(ownProps)
  let onCancel
```

webapp/src/components/AcceptBidAssetPage/AcceptBidAssetPage.container.js
```diff
@@ -0,0 +1,72 @@
...
+
+  switch (ownProps.assetType) {
+    case ASSET_TYPES.parcel: {
+      const { x, y } = getMatchParamsCoordinates(ownProps)
+      onCancel = () => dispatch(push(locations.parcelDetail(x, y)))
+      break
+    }
+    case ASSET_TYPES.estate: {
+      const { id } = getMatchParams(ownProps)
```
nicosantangelo: Maybe it's better to call this once and use `matchParams.id` here

webapp/src/modules/wallet/utils.js
```diff
@@ -23,11 +25,19 @@ export function getWalletSagaOptions() {
       ),
       new Marketplace(env.get('REACT_APP_MARKETPLACE_CONTRACT_ADDRESS')),
       new EstateRegistry(env.get('REACT_APP_ESTATE_REGISTRY_CONTRACT_ADDRESS')),
+      new ERC721Bid(env.get('REACT_APP_ERC721_BID_CONTRACT_ADDRESS')),
```
nicosantangelo: Why is this here if why are adding it conditionally below?

Also, do you think it might look nicer if we extract this like we did with Mortgage and Auction?

I know it's only one contract but we can keep a single `return` (instead of `const options`) and once bid goes live it's just a matter of replacing the method with this line here.

And, the method is really simple
```
function getBidContracts() {
  return isFFEnabled() ? [contract] : []
}

File: webapp/src/modules/wallet/sagas.js [hunk 44]
```diff
@@ -55,6 +56,10 @@ function* handleConnectWalletSuccess(action) {
     }
   }
 
+  if (isFeatureEnabled('BIDS')) {
+    authorization.allowances['ERC721Bid'] = ['MANAToken']
```
nicosantangelo: You can use a dot notation here if you want
```suggestion
    authorization.allowances.ERC721Bid = ['MANAToken']
```

webapp/src/modules/asset/utils.js
```diff
@@ -0,0 +1,42 @@
...
+export function getNFTAddressByType(type) {
+  if (type === ASSET_TYPES.parcel) {
+    return getContractAddress('LANDRegistry')
+  } else if (type === ASSET_TYPES.estate) {
+    return getContractAddress('EstateRegistry')
+  }
+  throw Error('invalid asset type')
+}
+
+export async function buildAsset(assetId, assetType, estates) {
```
nicosantangelo: I don't like this the last argument. It's conditional to assetType, and it tomorrow a new assetType appears that needs part of the state like estates does, we don't have a place to add it. Also this is forcing the caller to **always** instantiate all estates before building a random asset.
It made sense before to read the current state and build the asset from there.

A solution might be exporting this helper directly from `/modules/asset/sagas`, just using store.getState(), or something else

cc: @cazala

webapp/src/modules/asset/utils.js
```diff
@@ -0,0 +1,42 @@
...
+    asset = {
+      id: assetId,
+      data: {
+        name: estate.data.name,
+        parcels: estate.data.parcels
+      }
+    }
+  }
+
+  // TODO: asset.id should be asset.token_id
```
nicosantangelo: This TODO is no longer applicable

webapp/src/modules/bid/selectors.js
```diff
@@ -0,0 +1,70 @@
...
+export const getError = state => getState(state).error
+
+export const isBidIdle = state => isLoadingType(getLoading(state), BID_REQUEST)
+export const isCancelIdle = state =>
+  isLoadingType(getLoading(state), CANCEL_BID_REQUEST)
+
+export const getBidById = (state, bidId, assetType, assetId) => {
+  const bids = getData(state)
+  const bid = bids[bidId]
+  return bid && bid.asset_id === assetId && bid.asset_type === assetType
```
nicosantangelo: I don't get this check. We're already supplying this method with and bidId, why check the assetType/assetId too?
Maybe that should be done by the caller if necessary?

webapp/src/modules/bid/selectors.js
```diff
@@ -0,0 +1,70 @@
...
+    const bidId = Object.keys(bids).filter(bidId => {
+      const bid = bids[bidId]
+      return (
+        bid &&
+        bid.asset_id === assetId &&
+        bid.asset_type === assetType &&
+        bid.bidder === userAddress
+      )
+    })[0]
+    return bidId ? bids[bidId] : null
```
nicosantangelo: I don't understand this selector

- Why the `filter` if we're getting the first result?
- Why get only the `bidId` and then re-access the `bids[bidId]`
- Are you trying to do a `find`?

webapp/src/modules/bid/selectors.js
```diff
@@ -0,0 +1,70 @@
...
+  })
+
+export const getBidsByAssetFactory = (isOwner, assetType, assetId) =>
```
nicosantangelo: This method name is confusing to me. The selector above is called `..ByAssetId..` and receives an assetType an assetId, but this one is called `..byAsset..` and receives `isOwner`?

Also, having a boolean as the first argument is never a good idea, I had to come back here after stumbling upon `getBidsByAssetFactory(true, ASSET_TYPES.estate, id)` to see what true meant

webapp/src/modules/bid/selectors.js
```diff
@@ -0,0 +1,70 @@
...
+  createSelector(getAddress, getData, (userAddress, allBids) =>
+    Object.keys(allBids).reduce((bids, bidId) => {
+      const bid = allBids[bidId]
+      if (
+        bid.asset_type === assetType &&
+        bid.asset_id === assetId &&
+        bid.status === LISTING_STATUS.open &&
+        ((isOwner && bid.seller === userAddress) ||
+          (!isOwner && bid.bidder === userAddress))
```
nicosantangelo: You think this condition

```js
((isOwner && bid.seller === userAddress) ||
          (!isOwner && bid.bidder === userAddress))
```

might have a name and be extracted?

webapp/src/modules/bid/selectors.js
```diff
@@ -0,0 +1,70 @@
...
+      if (
+        bid.asset_type === asset.type &&
+        bid.asset_id === asset.id &&
+        bid.status === LISTING_STATUS.sold
+      ) {
+        return [...bids, bid]
+      }
+      return bids
+    }, [])
+  )
```
nicosantangelo: **File comment 1**

I see that this check 

```js
    bid.asset_type === assetType &&
    bid.asset_id === assetId
```

is fairly common. You think it would make sense to extract and name it?

nicosantangelo: **File comment 2**

⚠️ I see you're returning memoized selectors, but I think you're using them on mapStateToProps as " normal" selectors thus waisting the memo, i.e. `AcceptBidAssetPage.container.js`. We should either remove the createSelector or re-write the container

webapp/src/modules/bid/utils.js
```diff
@@ -0,0 +1,21 @@
+export const getUniqueAssetBidIds = (bidIds, allBids) => {
+  const assetIds = {}
+  return bidIds.reduce((ids, id) => {
+    const bid = allBids[id]
+
+    if (bid) {
```
nicosantangelo: Why wouldn't it be defined?

webapp/src/modules/bid/utils.js
```diff
@@ -0,0 +1,21 @@
...
+      const assetKey = `${bid.asset_type}_${bid.asset_id}`
```
nicosantangelo: Why not just using the bidId instead of an arbitrary string?

webapp/src/modules/bid/utils.js
```diff
@@ -0,0 +1,21 @@
...
+      if (!assetIds[assetKey]) {
+        assetIds[assetKey] = true
+        return [...ids, id]
+      }
+    }
+    return ids
+  }, [])
```
nicosantangelo: I think this method is trying to replicate a `Set` by using an object as cache. I think it'll look waay simpler just using the language structure

webapp/src/components/AcceptBidAssetPage/AcceptBidEstatePage/AcceptBidEstatePage.js
```diff
@@ -0,0 +1,84 @@
...
+import { Loader } from 'semantic-ui-react'
+import { t, T } from '@dapps/modules/translation/utils'
+
+import Estate from 'components/Estate'
+import EstateModal from 'components/EditEstatePage/EditEstateMetadata/EstateModal'
+import EstateName from 'components/EstateName'
+import Mana from 'components/Mana'
+import { walletType, bidType } from 'components/types'
+
+// import './BuyEstatePage.css'
```
nicosantangelo: Whops

webapp/src/components/AcceptBidAssetPage/AcceptBidEstatePage/AcceptBidEstatePage.js
```diff
@@ -0,0 +1,84 @@
...
+                  asset_type: t('name.estate')
+                })}
+                subtitle={
+                  <T
+                    id="asset_accept_bid.about_to_accept_bid"
+                    values={{
+                      name: <EstateName estate={estate} />,
+                      price: (
+                        <React.Fragment>
+                          {t('global.for')}&nbsp;&nbsp;
```
nicosantangelo: We might want to revisit this as it's probably not translatable to non-romantic languages

webapp/src/components/AcceptBidAssetPage/AcceptBidEstatePage/AcceptBidEstatePage.js
```diff
@@ -0,0 +1,84 @@
...
+                          <span>
```
nicosantangelo: Is the `span` necessary?

webapp/src/components/BidAssetPage/BidEstatePage/BidEstatePage.js
```diff
@@ -0,0 +1,124 @@
...
+    const bidIsOpen = isOpen(bid)
+
+    if (isLoading) {
+      return this.renderLoading()
+    }
+
+    return (
+      <Estate id={id} ownerNotAllowed>
+        {estate => {
+          const isERC721BidAllowed =
```
cazala: Suggestion: use implicit return to avoid nested `return` statements like the one below, and extract the variable declarations like this one to a class method that can be called from within the implicit return, like `this.isAllowed()` and then having 

```js
isAllowed = () => {
  return this.props.authorization && this.props.authorization.allowances.ERC721Bid.MANAToken
}
```

src/sanitize/sanitizeAssets.js
nachomazzara: 👍 Removed.  I'm not using bids inside any asset

webapp/src/components/Asset/Asset.js
nachomazzara: In order to re-fetch the asset when showing the detail

webapp/src/components/BidAssetPage/BidAssetForm.js
nachomazzara: good Catch

webapp/src/components/CancelBidAssetPage/CancelBidAssetPage.js
```diff
@@ -0,0 +1,43 @@
...
+  componentWillReceiveProps(nextProps) {
+    const { isConnected, isLoading, onFetchBidByAsset, bid } = nextProps
+
+    if (!isConnected || isLoading || bid || this.bidFetched) {
+      return
+    }
+    this.bidFetched = true
+
+    onFetchBidByAsset()
+  }
```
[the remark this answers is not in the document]
nachomazzara: Here it is different because we need the user to be connected before fetching its bids for the required asset

webapp/src/components/ParcelDetailPage/ParcelDetailPage.js
```diff
@@ -84,12 +87,12 @@ export default class ParcelDetailPage extends React.PureComponent {
               />
             </Grid.Column>
           </Grid.Row>
-          {publication || isOwner ? (
+          {publication || isOwner || !isOwner ? (
```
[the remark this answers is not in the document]
nachomazzara: 🤦‍♂️

webapp/src/modules/bid/actions.js
```diff
@@ -0,0 +1,187 @@
...
+}
+
+export function bidSuccess(txHash, bid, asset) {
+  return {
+    type: BID_SUCCESS,
+    ...buildTransactionWithReceiptAction(txHash, {
+      price: bid.price,
+      bidder: bid.bidder,
+      expires_at: bid.expires_at,
+      ...asset
```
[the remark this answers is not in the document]
nachomazzara: Yep, I need to have at least that info to display the bid if you are still seeing the asset detail page

webapp/src/modules/bid/actions.js
```diff
@@ -0,0 +1,187 @@
...
+}
+
+export function acceptBidSuccess(txHash, bid, asset) {
+  return {
+    type: ACCEPT_BID_SUCCESS,
+    ...buildTransactionAction(txHash, {
+      bidId: bid.id,
+      price: bid.price,
+      bidder: bid.bidder,
+      ...asset
```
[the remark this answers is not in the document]
nachomazzara: Same answer

webapp/src/modules/bid/sagas.js
nachomazzara: My bad

webapp/src/modules/estates/reducer.js
nachomazzara: Implicit return?

webapp/src/modules/address/reducer.js
```diff
@@ -118,10 +130,53 @@ export function addressReducer(state = INITIAL_STATE, action) {
...
+        ...(addressData.parcel_ids || []),
+        ...toAddressParcelIds(
+          bids
+            .filter(
+              ({ asset_type, seller }) =>
+                ASSET_TYPES.parcel === asset_type && seller === action.address
+            )
+            .map(({ asset_id }) => asset_id)
+        )
+      ])
```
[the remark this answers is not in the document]
nachomazzara: 🤷‍♂️

src/Listing/Bid/Bid.model.js
nachomazzara: Right, I will change the order and re-visit names

PR #835, opened 2019-01-29
feat: sanitize actions and state based on .env

nachomazzara:
New webapp envs

```
# Redux

REACT_APP_REDUX_ACTION_ONLY_TYPE=
REACT_APP_REDUX_STORE_KEYS=
```

You can set something like:

```
REACT_APP_REDUX_ACTION_ONLY_TYPE=1
REACT_APP_REDUX_STORE_KEYS=parcels,address
```

And see only: 

screen shot 2019-01-29 at 18 08 37
screen shot 2019-01-29 at 18 08 41


webapp/src/store.js
```diff
@@ -20,7 +20,25 @@ import { etherscan } from 'lib/EtherscanAPI'
...
+          ? storeKeys.split(',').reduce((sanitizedState, key) => {
+              const value = state[key]
+              if (value) {
+                return { ...sanitizedState, [key]: value }
+              }
+              return sanitizedState
+            }, {})
+          : state
+    })
+  : compose
```
nicosantangelo: Maybe we should do this only when `env.isDevelopment()` ?

webapp/.env.example
```diff
@@ -72,3 +72,8 @@ REACT_APP_MKR_MAX_CONVERSION_AMOUNT=
 REACT_APP_SNT_MAX_CONVERSION_AMOUNT=
 REACT_APP_ZIL_MAX_CONVERSION_AMOUNT=
 REACT_APP_RCN_MAX_CONVERSION_AMOUNT=
+
+# Redux
+
+REACT_APP_REDUX_ACTION_ONLY_TYPE=
```
nicosantangelo: I think this should be called something like:

`REACT_APP_REDUX_ACTION_ONLY_DISPLAY_TYPE`
`REACT_APP_REDUX_ACTION_ONLY_SHOW_TYPE`
`REACT_APP_REDUX_ACTION_SKIP_PAYLOAD`

I thought it was a flag for letting only some types through

webapp/.env.example
```diff
@@ -72,3 +72,8 @@ REACT_APP_MKR_MAX_CONVERSION_AMOUNT=
...
+REACT_APP_REDUX_STORE_KEYS=
```
nicosantangelo: Regardless of the comment above, maybe we should point out that these env variables are for the redux devtools extension **only**, to avoid confusion.

```
REACT_APP_DEVTOOLS_EXTENSION_ACTION_ONLY_TYPE
REACT_APP_DEVTOOLS_EXTENSION_STORE_KEYS
```

webapp/src/store.js
nachomazzara: Hmmm, we can avoid both sanitizers keeping the env empty. Do you think that is still useful to check if the environment is development??

webapp/.env.example
nachomazzara: Agree with adding `_DEVTOOLS_EXTENSION_`

But, how do you think that would be the use case for: 

```
REACT_APP_REDUX_ACTION_ONLY_DISPLAY_TYPE
REACT_APP_REDUX_ACTION_ONLY_SHOW_TYPE
REACT_APP_REDUX_ACTION_SKIP_PAYLOAD
```

PR #845, opened 2019-02-11
feat: new logo

cazala:
I had to disable the loading logo because the current gif has a white background, until we get a transparent one.


webapp/src/images/logo.svg
```diff
@@ -1,29 +1,40 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<svg width="52px" height="52px" viewBox="0 0 52 52" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
-    <!-- Generator: Sketch 47.1 (45422) - http://www.bohemiancoding.com/sketch -->
-    <title>logo</title>
+<svg width="36px" height="36px" viewBox="0 0 36 36" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <!-- Generator: Sketch 53 (72520) - https://sketchapp.com -->
+    <title>36</title>
```
nachomazzara: Is this title ok? Can we change it to `Decentraland`?

cazala: that's the version of the logo, it's not shown to the user

nachomazzara: Talked to @cazala. It is ok.

PR #848, opened 2019-02-14
Feat: bid detail

nachomazzara:
Bid Details

screen shot 2019-02-14 at 13 08 10
screen shot 2019-02-14 at 13 08 00

### DIFF
--- a/webapp/src/components/Bid/Bid.css
+++ b/webapp/src/components/Bid/Bid.css
@@ -4,6 +4,7 @@
   width: 100%;
   color: white;
   font-size: 16px;
+  margin: 0;
 }
 
 .ui.grid.stackable.grid.Bid:not(.showAssetDetail) {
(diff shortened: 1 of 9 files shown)


webapp/src/components/Bid/Bid.css
```diff
@@ -76,21 +89,20 @@
   margin-left: 15px !important;
 }
 
-@media(min-width: 758px) and (max-width: 992px) {
-  .Bid .row h3 {
-    font-size: 13px;
+@media(max-width: 768px) { 
+  .Bid .row .column.asset-details a{
```
nicosantangelo:
```suggestion
  .Bid .row .column.asset-details a {
```

webapp/src/components/Bid/Bid.js
```diff
@@ -134,72 +133,91 @@ export default class Bid extends React.PureComponent {
...
+                  amount={bid.price}
+                  size={15}
+                  scale={1}
+                  className="mortgage-amount-icon"
+                />
+              </Grid.Column>
+              <Grid.Column
+                computer={showAssetDetail ? 4 : 3}
+                tablet={showAssetDetail ? 4 : 3}
+                mobile={showAssetDetail ? 8 : 16}
```
nicosantangelo: This seems to be repeated thrice, maybe we could do something like:

```js
const sizeByResolution = {
  computer: showAssetDetail ? 4 : 3
  tablet: showAssetDetail ? 4 : 3
  mobile: showAssetDetail ? 8 : 16
}
```

and then

```jsx
<Component
  {...sizeByResolution}
 />
```

nachomazzara: You are right! that is better

PR #857, opened 2019-02-22
Feat: add operator to assets

### LINKED ISSUE #858: Add operators to assets & accounts
-  Add `operator` column to assets 
- A new relationship for `approvedForAll`
- A new API endpoint to return whether an account is allowed to update asset content

nachomazzara:
New version of decentraland-eth needed

Depends on decentraland/decentraland-eth#56 and decentraland/decentraland-eth#57

Close #858


src/Authorization/Authorization.router.js
```diff
@@ -0,0 +1,118 @@
...
+      x = unsafeParseInt(server.extractFromReq(req, 'x'))
+    } catch (e) {
+      throw new Error('Invalid coordinate "x" must be an integer')
+    }
+
+    try {
+      y = unsafeParseInt(server.extractFromReq(req, 'y'))
+    } catch (e) {
+      throw new Error('Invalid coordinate "y" must be an integer')
+    }
```
nicosantangelo: We should use `ReqQueryParams` here which abstracts `unsafeParseInt` in `getInteger`

`const x = reqQueryParams.getInteger('x')`

src/Authorization/Authorization.router.js
```diff
@@ -0,0 +1,118 @@
...
+        tokenAddress = env.get('LAND_REGISTRY_CONTRACT_ADDRESS')
+        break
+      }
+      case ASSET_TYPES.estate: {
+        tokenAddress = env.get('ESTATE_REGISTRY_CONTRACT_ADDRESS')
+        break
+      }
+      default:
+        throw new Error(`The assetType ${assetType} is invalid`)
+    }
```
nicosantangelo: Two things about this one:

- I think we should use contractAddresses here instead of `env.get` as we do in the rest of the codebase...except Estate.modal, but we need to change that one
- *Not important*: You think it's good idea if we add a helper for this? I _think_ we needed it elsewhere

monitor/reducers/parcelReducer.js
```diff
@@ -29,10 +31,15 @@ export async function parcelReducer(event) {
...
+  let parcelId
 
-  const parcelId = await getParcelIdFromEvent(event)
-  if (!parcelId) return log.info(`[${name}] Invalid Parcel Id`)
+  if (!omitParcelId(event.name)) {
+    parcelId = await getParcelIdFromEvent(event)
+    if (!parcelId && omitParcelId(event.name)) {
+      return log.info(`[${name}] Invalid Parcel Id`)
+    }
+  }
```
nicosantangelo: Huh I don't get this part.

- If `omitParcelId(event.name)` is `true` this won't run and, `parcelId` will be undefined below
- If `omitParcelId(event.name)` is `false` you don't need to check it again on line 39

monitor/reducers/estateReducer.js
```diff
@@ -166,6 +169,46 @@ async function reduceEstateRegistry(event) {
...
+      const _approved = event.args._approved === 'true'
+
+      try {
+        log.info(
+          `[${name}] ${_owner} ${
+            _approved ? 'set' : 'remove'
+          } ${_operator} as approved for all`
+        )
+        if (_approved) {
+          await Approval.insertApproval(address, _owner, _operator)
```
nicosantangelo: check insertApproval comment below

monitor/reducers/parcelReducer.js
nachomazzara: Good catch! I need to remove `omitParcelId` from the second if

monitor/reducers/utils.js
```diff
@@ -45,3 +47,7 @@ export async function getParcelIdFromEvent(event) {
   const { assetId, landId, _landId, _tokenId } = event.args
   return Parcel.decodeTokenId(assetId || landId || _landId || _tokenId)
 }
+
+export function omitParcelId(name) {
+  return [eventNames.ApprovalForAll].includes(name)
```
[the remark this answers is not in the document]
nachomazzara: We can have more events that don't need to check the `parcelId` as `approvalForAll`. I wanted to leave it ready for that, the performance is imperceptible in this case.

src/Authorization/Authorization.router.js
nachomazzara: The helper will work.  We should need where we use `contractAddresses` with a switch by `ASSET_TYPE` ([issue created](https://github.com/decentraland/marketplace/issues/875))

nachomazzara: I should revert it because when we start the server we are not connecting to the eth which is need for `ethereum.js` to work

monitor/reducers/utils.js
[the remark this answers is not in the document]
nicosantangelo: Allright. I'm not really an advocate for "leaving it for the future programmer to figure it out", if that's the case I prefer a comment and the correct implementation for the current case (it's not about performance). Let's leave it like this no worries.

On the other hand, the name is not correct (forgot to add it in the first comment, sorry). The method is not omitting anything and there's not connection to a `parcelId`.
I think maybe `shouldOmitEvent` makes more sense

src/Approval/Approval.model.js
```diff
@@ -0,0 +1,26 @@
...
+    )
+  }
+
+  static async isApprovedForAll(tokenAddress, owner, operator) {
+    const res = await this.findOne({
+      token_address: tokenAddress,
+      owner,
+      operator
+    })
+    return res !== undefined
```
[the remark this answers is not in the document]
nicosantangelo: ☝️

monitor/reducers/parcelReducer.js
```diff
@@ -87,12 +94,66 @@ async function reduceLANDRegistry(event) {
...
+          } ${operator} as approved for all`
+        )
+        if (authorized) {
+          await Approval.insertApproval(address, holder, operator)
+        } else {
+          await Approval.delete({
+            token_address: address,
+            owner: holder.toLowerCase(),
+            operator: operator.toLowerCase()
+          })
```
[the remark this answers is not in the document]
nicosantangelo: The new changes use the `toLowercase` correctly, but they're still using `insertApproval` which is not necessary and can be deleted

src/Authorization/Authorization.router.js
nicosantangelo: Ah that's a shame. Maybe we can call `loadContracts` from the server instead (it doesn't connect to a node). We'll leave it to the other issue

monitor/reducers/parcelReducer.js
nicosantangelo: Still my first point stands, what happens when the first if it's not fulfilled:

```
if (!omitParcelId(event.name)) {                    // <==== This one
  parcelId = await getParcelIdFromEvent(event)
  if (!parcelId) {
    return log.info(`[${name}] Invalid Parcel Id`)
  }
}
```

parcelId will be `undefined` moving forward. Isn't that a problem?

monitor/reducers/estateReducer.js
nicosantangelo: ☝️

monitor/reducers/parcelReducer.js
nachomazzara: Not a problem

monitor/reducers/utils.js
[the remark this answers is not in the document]
nachomazzara: It is not an event omit, it is just an omission of a specific argument of the event.

PR #885, opened 2019-03-26
Feat: archive bids

nachomazzara:
Screen Shot 2019-03-26 at 13 22 23
Screen Shot 2019-03-26 at 13 22 14


webapp/src/components/ParcelDetailPage/ParcelDetailPage.container.js
```diff
@@ -13,14 +14,16 @@ import ParcelDetailPage from './ParcelDetailPage'
 const mapState = (state, ownProps) => {
   const parcel = ownProps.asset
   const getParcelMortgage = getParcelMortgageFactory(parcel.x, parcel.y)
+  const bids = getWalletBidsByAsset(state, ownProps.asset, ASSET_TYPES.parcel)
+  const archivedBids = getArchivedBids(state)
```
nicosantangelo: **To discuss**

Maybe this: 

```
  const bids = getWalletBidsByAsset(state, ownProps.asset, ASSET_TYPES.parcel)
  const archivedBids = getArchivedBids(state)
  bids.filter(bid => !archivedBids[bid.id])
```

can be a selector in archiveBid?

getWalletArchivedBidsByAsset

webapp/src/components/ProfilePage/ProfilePage.container.js
```diff
@@ -42,6 +43,13 @@ const mapState = (state, { location, match }) => {
   }
 
   const publishedAssets = publishedParcels.concat(publishedEstates)
+  const archivedBids = getArchivedBids(state)
+
+  if (tab === PROFILE_PAGE_TABS.archivebids) {
+    bids = bids.filter(bid => archivedBids[bid.id])
+  } else {
+    bids = bids.filter(bid => !archivedBids[bid.id])
+  }
```
nicosantangelo: shouldn't this be a part of the switch below?

the true condition should be on `case PROFILE_PAGE_TABS.archivebids:`

and the false on `case PROFILE_PAGE_TABS.bids:`

I don't think it'd be a problem with the tab count because it's changing at the moment (it shows the archived/unarchived count)

webapp/src/components/ProfilePage/ProfilePage.js
```diff
@@ -233,6 +263,13 @@ export default class ProfilePage extends React.PureComponent {
     this.isActive(PROFILE_PAGE_TABS.bids) ||
     this.isActive(PROFILE_PAGE_TABS.mortgages)
 
+  isBidActive() {
```
nicosantangelo: maybe we can follow the way the method above is written

```javascript
  onlyOwnerCanAccess = () =>
    this.isActive(PROFILE_PAGE_TABS.bids) ||
    this.isActive(PROFILE_PAGE_TABS.mortgages)

  isBidActive = () =>
    this.isActive(PROFILE_PAGE_TABS.bids) ||
    this.isActive(PROFILE_PAGE_TABS.archivebids)
```

webapp/src/components/ProfilePage/ProfilePage.js
```diff
@@ -302,11 +339,11 @@ export default class ProfilePage extends React.PureComponent {
                 {isFeatureEnabled('BIDS') && (
                   <Menu.Item
                     name={PROFILE_PAGE_TABS.bids}
-                    active={this.isActive(PROFILE_PAGE_TABS.bids)}
+                    active={this.isBidActive()}
                     onClick={this.handleItemClick}
                   >
                     {t('global.bids')}
-                    {this.renderBadge(bids, PROFILE_PAGE_TABS.bids)}
+                    {this.renderBidBadge(bids)}
```
nicosantangelo: I don't think this is bad per-se, but instead of adding a new method what about doing:

```javascript
{this.renderBadge(
  bids,
  this.isActive(PROFILE_PAGE_TABS.archivebids)
    ? PROFILE_PAGE_TABS.archivebids
    : PROFILE_PAGE_TABS.bids
)}
```

webapp/src/modules/archivedBid/reducer.js
```diff
@@ -0,0 +1,30 @@
...
+          [bid.id]: true
+        }
+      }
+    }
+    case UNARCHIVE_BID: {
+      const { bid } = action
+      return {
+        data: {
+          ...state.data,
+          [bid.id]: false
```
nicosantangelo: it might be useful, instead of storing true/false, having timestamp/undefined. So:

```
data: {
  [bid.id]: Date.now()
````

and

```
data: {
  ...utils.omit(state.data, [bid.id])
```

if you don't find the timestamp useful, I still think cleaning the state might be worth it

webapp/src/components/ProfilePage/ProfilePage.container.js
nachomazzara: If you place that as part of the switch, you will see a count with all the bids.

When you are in another tab (not `bid`), you want to see the number of bids that you will see once you enter in that tab. And, as archived ones are hidden, the correct number to show is the unarchive bids.

PR #893, opened 2019-03-29
chore: delay error logging to allow for progress bars

nicosantangelo:


sanity/doctors/Doctor.js
```diff
@@ -1,5 +1,15 @@
...
+  constructor() {
+    this.log = new Log(this.constructor.name)
+  }
+
   async diagnose(options) {
     throw new Error('Not implemented')
   }
+
+  logErrors(faultyAssets) {
+    this.log.error(faultyAssets.map(({ error }) => error).join('\n'))
```
nachomazzara: I know this way of looping and logging is way faster than 

`faultyAssets.forEach({errorr} => this.log.error(error))`

But if we have a lot of errors, more memory will be needed. 

What do you think of adding a comment saying that memory will be needed _in a linear way_ related to the count of errors

nicosantangelo: Huh, haven't thought of that you're right. I'll add the comment but if you think this might be a problem we can change it to 

```javascript
faultyAssets.forEach(({error}) => this.log.error(error))
```

nicosantangelo:
> What do you think of removing this.logErrors(faulty...) from specific doctors, but adding it here to be consistent in the way we use them. Also, we can add the option you wanted skip-logs and skip them easily without passing down options to the doctors.

At first glance, I think this is a great idea! adding logErrors in each doctor felt weird but I didn't come up with a better solution. I'll give it a try!

sanity/SanityActions.js
```diff
@@ -32,6 +32,10 @@ export class SanityActions {
 
       if (diagnoses.hasProblems()) {
         faultyAssets.push(...diagnoses.getFaultyAssets())
+
+        if (options.skipLogs) {
```
nachomazzara:
```suggestion
        if (!options.skipLogs) {
```

PR #894, opened 2019-03-29
chore: DRY code a bit between doctors

nicosantangelo:


sanity/doctors/EstateDoctor.js
```diff
@@ -152,23 +152,33 @@ export class EstateDiagnosis extends Diagnosis {
...
-          // Replay estate events only for update or creation
-          await this.replayEvents([event])
-        }
+
+      const events = await this.getEventsToReplay(estate.id)
+      await this.replayEvents(events)
+    }
+  }
+
+  async getEventsToReplay(estateId) {
```
nachomazzara: Should we make this method static? it is weird to see a  `new EstateDiognosis()` at ParcelDoctor.

Also, why not creating utils inside /doctors root with this method? We can add a doc comment like:

```
Get events to replay for a specific Estate and all of its parcels that were or are part of it
```

Naming ideas: 
- `getEventsToReplayByEstateId`
- `getEventsByEstateId`
- `getAssetEventsByEstateId`
- `getAssetEventsToReplayByEstateId`

nicosantangelo: > it is weird to see a new EstateDiognosis() at ParcelDoctor.

Why?

> Also, why not creating utils inside /doctors root with this method? We can add a doc comment like

to avoid having a bag of methods where we can keep them inside a coherent namespace. I have no idea what I can find inside a `utils` file

nachomazzara: Because the method will be the same for all DoctorDiagnosis instances as we do something related to the input (`estateId` )without using any property of the class on it. Moreover, It feels weird to initialize a class without arguments and only use a method instead of the class itself (that's why making it static will help).

I know that Parcel can use Estate helpers to do estates related stuff, but in this case, we are using a method which is treating things like an asset (Estate and Parcel).  

We can have an `AssetDiagnoisHelper.js` which can help with
```
to avoid having a bag of methods where we can keep them inside a coherent namespace. I have no idea what I can find inside a utils file, no?
```

Again, it just an ideas/opinion, you can leave as it is if you want. No even need to rename the method if we will keep it in the EstateDiagnosis.

abarmat: I think it's a good observation but for now I'll keep the pattern of Doctors creating Diagnosis as new classes. Though now it's an instantiation without any param it might be in the future. Let's evaluate if we want to refactor this later. 👍

PR #896, opened 2019-03-29
feat: support database reconnections

### LINKED ISSUE #874: App hangs when database connection is lost

nicosantangelo:
You can use `DATABASE_RECONNECTION_TIMEOUT=` to set the timeout between reconnections, in ms. 3s by default

Closes #874


src/database/database.js
```diff
@@ -9,13 +9,28 @@ export const database = {
...
+
     for (const callback of connectCallbacks) {
       callback()
     }
 
     return this
   },
 
+  async reconnect() {
+    const timeout = env.get('DATABASE_RECONNECTION_TIMEOUT', '3000')
```
nachomazzara: Is there any reason why we use `'3000'` instead of just `3000`?

nachomazzara:
> The database connection recovers successfully but after multiple times killing and restoring the db I get this error and it doesn't recover:
> (this sometimes happens after only one or two kill/restore)
> 
> ```
> (node:72431) UnhandledPromiseRejectionWarning: Error: This socket has been ended by the other party
>     at Socket.writeAfterFIN [as write] (net.js:396:12)
>     at Connection.end (/Users/abarmat/dev/decentraland/marketplace/node_modules/pg/lib/connection.js:338:22)
>     at Client.end (/Users/abarmat/dev/decentraland/marketplace/node_modules/pg/lib/client.js:512:21)
>     at Object._callee$ (/Users/abarmat/dev/decentraland/marketplace/src/database/database.js:13:19)
> ```

It seems that `this.client.end()` is trying to close an already closed session. Maybe we can check the [error from the end() callback](https://node-postgres.com/api/client#client-end-cb-err-error-gt-void-gt-void) and do whethever we need

src/database/database.js
nicosantangelo: I kept the habit from typescript, where `env.get('DATABASE_RECONNECTION_TIMEOUT', 3000)` fails because `env.get('DATABASE_RECONNECTION_TIMEOUT')` returns a string but the default value is a number

So, it's so the type is consistent. We can change it if you think it's best!

nicosantangelo:
Thank you both for the review, I'm checking the error!

nicosantangelo:
I can add a try catch like nacho said but I'm not being able to reproduce the error. I tried killing and restarting the database 10 times with requests in the middle and it still works.

How are you killing the service @abarmat ?

src/database/database.js
nachomazzara: I'm just asking because we have an int in all the fallbacks values which are expecting a number. I think that it would be better to use `3000` to be aligned with the rest but is up to you 😄.

PR #898, opened 2019-03-31
chore: update run sanity script

abarmat:


nachomazzara:
Awesome :) 

What do you think of waiting for #899?

nicosantangelo:
I think we can merge this one as #899 shouldn't change the way we run the script, or do you mean waiting for the deploy?

abarmat:
👍 I'm running currently with that config, we could try tweaking it better after. Thank you for the review!

PR #911, opened 2019-04-23
feat: sanitize assets operators

nachomazzara:
- Sanitize Asset `updateOperators` and `operators`

### DIFF
--- a/sanity/doctors/EstateDoctor.js
+++ b/sanity/doctors/EstateDoctor.js
@@ -123,7 +123,8 @@ export class EstateDoctor extends Doctor {
 
   isOperatorMismatch(currentOperator, operator) {
     return (
-      (!operator && !Contract.isEmptyAddress(currentOperator)) ||
+      (Contract.isEmptyAddress(operator) &&
+        !Contract.isEmptyAddress(currentOperator)) ||
       (operator && operator !== currentOperator)
     )
   }
--- a/sanity/doctors/EstateDoctor.js
+++ b/sanity/doctors/EstateDoctor.js
@@ -44,41 +44,57 @@ export class EstateDoctor extends Doctor {
     const {
       id,
       owner,
+      operator,
+      update_operator,
       data: { parcels }
     } = estate
-    let error = ''
 
     const estateRegistry = eth.getContract('EstateRegistry')
-    const currentOwner = await estateRegistry.ownerOf(estate.id)
+    const [
+      currentOwner,
+      currentOperator,
+      currentUpdateOperator
+    ] = await Promise.all([
+      estateRegistry.ownerOf(estate.id),
+      estateRegistry.getApproved(estate.id),
+      estateRegistry.updateOperator(estate.id)
+    ])
 
     if (owner !== currentOwner) {
-      error = `Mismatch: owner of '${id}' is '${owner}' on the DB and '${currentOwner}' in blockchain`
-    } else {
-      const estateSize = await estateRegistry.getEstateSize(estate.id)
-      let currentParcels = []
-
-      let index = 0
-      await asyncBatch({
-        elements: new Array(estateSize.toNumber()),
-        callback: async sizeBatch => {
-          const promises = []
-          for (let i = 0; i < sizeBatch.length; i++) {
-            promises.push(this.buildCurrentEstateParcel(estate.id, index++))
-          }
-
-          const parcels = await Promise.all(promises)
-          currentParcels = [...currentParcels, ...parcels]
-        },
-        batchSize: env.get('BATCH_SIZE'),
-        logFormat: '',
-        retryAttempts: 20
-      })
-      if (!this.isEqualParcels(currentParcels, parcels)) {
-        error = `Parcels: db parcels for estate ${id} are different from blockchain`
-      }
+      return `Mismatch: owner of '${id}' is '${owner}' on the DB and '${currentOwner}' in blockchain`
     }
 
-    return error
+    if (this.isOperatorMismatch(currentOperator, operator)) {
+      return `Mismatch: operator of '${id}' is '${operator}' on the DB and '${currentOperator}' in blockchain`
+    }
+
+    if (this.isOperatorMismatch(currentUpdateOperator, update_operator)) {
+      return `Mismatch: update operator of '${id}' is '${update_operator}' on the DB and '${currentUpdateOperator}' in blockchain`
+    }
+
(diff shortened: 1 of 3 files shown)


sanity/doctors/EstateDoctor.js
```diff
@@ -44,41 +44,57 @@ export class EstateDoctor extends Doctor {
...
+        currentParcels = [...currentParcels, ...parcels]
+      },
+      batchSize: env.get('BATCH_SIZE'),
+      logFormat: '',
+      retryAttempts: 20
+    })
+    if (!this.isEqualParcels(currentParcels, parcels)) {
+      return `Parcels: db parcels for estate ${id} are different from blockchain`
+    }
+    return null
```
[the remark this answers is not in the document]
nicosantangelo: Also, I feel this block of code should be it's own method:

```javascript
    const estateSize = await estateRegistry.getEstateSize(estate.id)
    let currentParcels = []

    let index = 0
    await asyncBatch({
      elements: new Array(estateSize.toNumber()),
      callback: async sizeBatch => {
        const promises = []
        for (let i = 0; i < sizeBatch.length; i++) {
          promises.push(this.buildCurrentEstateParcel(estate.id, index++))
        }

        const parcels = await Promise.all(promises)
        currentParcels = [...currentParcels, ...parcels]
      },
      batchSize: env.get('BATCH_SIZE'),
      logFormat: '',
      retryAttempts: 20
    })
    if (!this.isEqualParcels(currentParcels, parcels)) {
      return true
    }
    return false
```

it's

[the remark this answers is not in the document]
nicosantangelo: Also, I feel this block of code should be it's own method:

```javascript
    const estateSize = await estateRegistry.getEstateSize(estate.id)
    let currentParcels = []

    let index = 0
    await asyncBatch({
      elements: new Array(estateSize.toNumber()),
      callback: async sizeBatch => {
        const promises = []
        for (let i = 0; i < sizeBatch.length; i++) {
          promises.push(this.buildCurrentEstateParcel(estate.id, index++))
        }

        const parcels = await Promise.all(promises)
        currentParcels = [...currentParcels, ...parcels]
      },
      batchSize: env.get('BATCH_SIZE'),
      logFormat: '',
      retryAttempts: 20
    })
    if (!this.isEqualParcels(currentParcels, parcels)) {
      return true
    }
    return false
```

it's self-contained enough and distracts from the rest of the method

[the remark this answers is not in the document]
nicosantangelo: Also, I feel this block of code should be it's own method:

```javascript
    const estateSize = await estateRegistry.getEstateSize(estate.id)
    let currentParcels = []

    let index = 0
    await asyncBatch({
      elements: new Array(estateSize.toNumber()),
      callback: async sizeBatch => {
        const promises = []
        for (let i = 0; i < sizeBatch.length; i++) {
          promises.push(this.buildCurrentEstateParcel(estate.id, index++))
        }

        const parcels = await Promise.all(promises)
        currentParcels = [...currentParcels, ...parcels]
      },
      batchSize: env.get('BATCH_SIZE'),
      logFormat: '',
      retryAttempts: 20
    })
    if (!this.isEqualParcels(currentParcels, parcels)) {
      return true
    }
    return false
```

it's self contained and distracts from the main method

sanity/doctors/EstateDoctor.js
```diff
@@ -44,41 +44,57 @@ export class EstateDoctor extends Doctor {
     const {
       id,
       owner,
+      operator,
+      update_operator,
       data: { parcels }
```
nicosantangelo: I know this is not from this PR but the rest of the codebase tends to steer away from this double destructuring. I looked around and it's the only place I found where we're using it. If you don't mind, can we split it into two lines?

sanity/doctors/EstateDoctor.js
```diff
@@ -99,6 +120,13 @@ export class EstateDoctor extends Doctor {
 
     return leftIds.every((parcelId, index) => parcelId === rightIds[index])
   }
+
+  isOperatorMismatch(currentOperator, operator) {
+    return (
+      (!operator && !Contract.isEmptyAddress(currentOperator)) ||
```
nicosantangelo: *Question*: wouldn't it be more accurate to do this?:

```suggestion
      (Contract.isEmptyAddress(operator) && !Contract.isEmptyAddress(currentOperator)) ||
```

because the caller could send anything as the first argument, even an empty address

sanity/doctors/ParcelDoctor.js
```diff
@@ -75,10 +92,10 @@ export class ParcelDoctor extends Doctor {
     return !!currentOwner && parcel.owner !== currentOwner
   }
 
-  isUpdateOperatorMismatch(currentUpdateOperator, parcel) {
+  isOperatorMismatch(currentOperator, operator) {
     return (
-      !!currentUpdateOperator &&
-      parcel.update_operator !== currentUpdateOperator
+      (!operator && !Contract.isEmptyAddress(currentOperator)) ||
+      (operator && operator !== currentOperator)
```
nicosantangelo: This seems to be the same method as `isOperatorMismatch` on EstateDoctor, I think we might need to avoid duplication here.

Some ideas:
- Have an `Address` class:
```javascript
new Address(currentOperator).equals(operator)
// or
new EthAddress(currentOperator).equals(operator)
```
which would also allow for some other abstractions regarding addresses
- The same can be accomplished by having a file with functions. A bit worse for autocompletion and discoverability but maybe less complex.
- We could use composition

```javascript
export class ParcelDoctor extends Doctor {
  constructor(...props) {
    super(...props)
    this.assetDoctor = new AssetDoctor(this)
  }

  method() {
    this.assetDoctor.isAddressMismatch()
    // or
    !this.assetDoctor.isSameAddress()
  }
}
// (...)

export class EstateDoctor extends Doctor {
  constructor(...props) {
    super(...props)
    this.assetDoctor = new AssetDoctor(this)
  }

  method() {
    this.assetDoctor.isAddressMismatch()
    // or
    !this.assetDoctor.isSameAddress()
  }
}
```

we can maybe reuse the `ethereum/` folder for the first two

nachomazzara: I like the idea but I think that it should be part of another PR. 

What do you think? I can create an issue

nicosantangelo: 🆗

PR #914, opened 2019-04-25
feat: index UpdateManager

nicosantangelo:
Continues #913


monitor/reducers/approvalReducer.js
```diff
@@ -0,0 +1,100 @@
...
+        owner: event.args.holder.toLowerCase(),
+        operator: event.args.operator.toLowerCase()
+      }
+      const isApproved = event.args.authorized === 'true'
+
+      await handleApproval(event, approval, isApproved)
+      break
+    }
+    case eventNames.UpdateManager: {
+      // operator and holder are inverted in the contact
```
nachomazzara: The code for LANDRegistry looks similar to the EstateRegistry. What do you think of having only one method for both? Same with `ApprovalForAll`

monitor/reducers/approvalReducer.js
```diff
@@ -0,0 +1,100 @@
...
+    default:
+      break
+  }
+}
+
+async function reduceEstateRegistry(event) {
+  const { name, address } = event
+
+  switch (name) {
+    case eventNames.ApprovalForAll: {
```
nachomazzara: `UpdateManager` event is missing here, right? With the refactor you won't need it

monitor/reducers/approvalReducer.js
```diff
@@ -0,0 +1,100 @@
...
+      break
+  }
+}
+
+async function reduceLANDRegistry(event) {
+  const { name, address } = event
+
+  switch (name) {
+    case eventNames.ApprovalForAll: {
+      // operator and holder are inverted in the contact
```
nachomazzara: Within the new contracts upgrade, this comment is out of date. We should remove it.

src/ethereum/index.js
```diff
@@ -69,6 +69,7 @@ export function loadContracts() {
         'Update',
         'Transfer',
         'UpdateOperator',
+        'UpdateManager',
         'Approval',
         'ApprovalForAll'
       ],
```
nachomazzara: I know it is not part of this PR, but `Transfer` now has 3 types with that version of dcl-eth.
Is it ok for you to update it here?

PR #927, opened 2019-05-27
Feat: sanitize mana approvals

nachomazzara:
Sanitize MANA Approvals


sanity/doctors/ApprovalDoctor.js
```diff
@@ -30,15 +31,18 @@ export class ApprovalDiagnosis extends Diagnosis {
   async prepare(fromBlock) {
     this.fromBlock = Number(fromBlock) || 0
 
+    // Delete approvalForAll, UpdateManager and Mana Approval blockchain events
     return Promise.all([
       BlockchainEvent.deleteByName(eventNames.ApprovalForAll, fromBlock),
-      BlockchainEvent.deleteByName(eventNames.UpdateManager, fromBlock)
+      BlockchainEvent.deleteByName(eventNames.UpdateManager, fromBlock),
+      Approval.deleteManaApprovalEvents(fromBlock)
```
nicosantangelo: I think this is a bit confusing. It took me a while to understand what was going on. The comment below says that the approvals are only on the BlockchainEvents side but then the `Approval.model` is the one that does the delete.

Finding the `SQL.raw(BlockchainEvent.tableName)` is not the easier thing in the world.

I'd propose adding something like this to BlockchainEvent:

```
  static deleteFromBlock(blockchainEvent, fromBlock) {
    const attributes = utils.pick(blockchainEvent) // from decentraland-commons
    const values = Object.values(attributes, this.columnNames)
    return this.query(
      SQL`DELETE FROM ${SQL.raw(this.tableName)}
        WHERE ${this.db.toAssignmentFields(attributes)}
        AND ${BlockchainEventQueries.fromBlock(fromBlock)}`,
      values
    )
  }
```

PR #928, opened 2019-05-28
feat: sanitize avatar names

nachomazzara:
Sanitize claimed names


src/Asset/Estate/Estate.queries.js
```diff
@@ -7,16 +7,16 @@ import { BlockchainEventQueries } from '../../BlockchainEvent'
...
 export const EstateQueries = Object.freeze({
   // prettier-ignore
-  estateHasParcels: (tableRef) => 
+  estateHasParcels: (tableRef) =>
     SQL`(${raw(tableRef)}.asset_type = ${ASSET_TYPES.estate}
     AND EXISTS(SELECT 1 FROM ${raw(Parcel.tableName)} AS p WHERE p.estate_id = ${raw(tableRef)}.asset_id)
     OR ${raw(tableRef)}.asset_type != ${ASSET_TYPES.estate})`,
   areEstateEvents: estateId => {
-    const address = env.get('ESTATE_REGISTRY_CONTRACT_ADDRESS')
+    const address = env.get('ESTATE_REGISTRY_CONTRACT_ADDRESS').toLowercase()
```
nicosantangelo: we can start replacing this calls to ethereum/index.js `contractAddresses`, where they're all lowercased!

nachomazzara: This specific query nowadays is used in the sanity, but when we rise up the server we don't connect to Ethereum.

In that case, `contractAddress` will be empty.

nicosantangelo: Dang it!

nicosantangelo: I think we should run `loadContracts()` regardless, because it's not really node-related. It's just a method that loads the different Ethereum structures in memory.

I'm not sure if we should just do it on the file itself (on import). We can leave it for another PR

PR #934, opened 2019-06-21
feat: api address authorizations

abarmat:


abarmat:
@nicosantangelo I addressed the feedback

PR #957, opened 2019-10-21
feat: transfer LANDs with receipt

nachomazzara:


scripts/manageLAND.js
```diff
@@ -173,7 +295,22 @@ async function transferLANDs(args) {
...
-    return transferLANDs(landBatchessToRetry, ...args.slice(1))
+    return transferManyLAND(landBatchessToRetry, ...args.slice(1))
+  }
+}
+
+async function checkAllowance(land) {
+  const account = eth.getAccount()
+  const LANDRegistryContract = eth.getContract('LANDRegistry')
+  if (
+    (await LANDRegistryContract.getApproved(land.id)).toLowerCase() !==
```
nicosantangelo: Maybe we can move the `await` to a line above this one? to make it a bit less convoluted

scripts/manageLAND.js
```diff
@@ -67,45 +114,130 @@ const manageMana = {
...
+      }),
+    { batchSize, txDelay }
+  )
+
+  log.info(`Sent ${recipients.length - txsToRetry.length} transactions`)
+
+  if (txsToRetry.length > 0 && shouldRetry) {
+    const recipientsToRetry = txsToRetry.map(tx => tx.data)
+    log.info(`Retrying on ${recipientsToRetry.length} recipients`)
+    return transferLANDs(recipientsToRetry, ...args.slice(1))
```
nicosantangelo: Sorry for the silly question, but shouln't `args` be an object? (without a `slice` method)

scripts/manageLAND.js
nicosantangelo: Not for this PR, but it'd be awesome to have a `.isSameAddress()` method somewhere too.
For the future™

nicosantangelo (commented):
LGTM except one comment about a `.slice()` method

scripts/manageLAND.js
nachomazzara: Wow! good catch!

